PR #25, opened 2014-09-08
Travis CI fixup

hugovk:
The Travis CI builds were failing.
- The example show_tracks.py was timing out because it was waiting for some stdin which never appeared. I changed the default, when no parameters are given, to use the example long_track_test.dat
- Python 3 was failing in many places because it used Python 2-style prints. Updated to use `from __future__ import print_function` so the same Python 3-style `print()`s work on both Py2 and Py3.
- Python 3 is still failing because it doesn't like the import in `__init__.py`, but I've allowed Python 3 builds to fail for the time being (`allow_failures` in .travis.yml).
- Added a Travis CI build badge to the README so broken code can be discovered sooner.
- `flake8` for the modified files.

Please can you enable Travis CI and Coveralls for your account? 

Both are free, you log in with your GitHub ID:
- https://travis-ci.org/profile/
- https://coveralls.io/repos/new

Also, pull requests will be built so you can see the status before merging in.

Thanks!


plamere:
hey hugovk - My 2017 resolution (among others) is to stay on top of spotipy PRs.  This particular PR is the oldest one still pending.  I think that it is no longer appropriate - but I'm not a python 3 or travis user so I don't know for sure. If you are still interested in spotipy I'd appreciate it if you'd see if there's anything lacking in the latest version to make travis and python 3 folks happy.

hugovk:
Happy new year! Sounds like a good resolution!

As a minimum, I'd recommend you enable Travis CI for this repo, that way whenever someone sends a pull request, the tests will be automatically run (on clean environments) and you'll be able to see if their PR breaks some existing code too. Then you know before even merging whether to ask them to fix something.

It'd be handy if you could enable Coveralls too.

I'll have a check and see if anything needs updating for Python 3 and update this PR or create a new one.

plamere:
cleaning up old and probably stale PRs

hugovk:
Yep, tests are now failing with " No token provided" or hanging and timing out the build.
https://travis-ci.org/hugovk/spotipy/jobs/276583952

plamere:
yes, those tests fail, AFAICT because the spotify web api now requires authentication for all methods. Will be upgrading the tests (and the associated docs soon)

PR #28, opened 2014-09-18
Support for Python 3

joohoi:
With these changes Spotipy works with both Python 2.7 and Python 3.4, even though the examples still remain untested. All existing tests pass on both platforms.


joohoi:
Closed in favor of the more recent one #56

Issue #29, opened 2014-10-05
window not launching for oAuth


vyaly:
https://github.com/plamere/spotipy/tree/master/examples/
I tried to run user_playlists_contents.py but it doesn't launch the window.
I have set the proper credentials as per the instructions.

I am developing a webapp with Flask and Python and want the user to login via spotify on my home page. 

Please help!!!

PR #30, opened 2014-10-08
Fixed browser opening issue

sepehr:
There is an i-donno-what issue when `spotipy.uitl.prompt_for_user_token()` tries to open the browser using the `subprocess.call`. #29 issue might be related to this.  

This pull request intends to fix this by utilizing `webbrowser.open` instead. It also works around the unnecessary verbose output of `webbrowser.open` and programmatically redirects it to the `/dev/null`.  

The code is tested on both OSX and Linux.

### DIFF
--- a/spotipy/util.py
+++ b/spotipy/util.py
@@ -67,17 +67,20 @@ def prompt_for_user_token(username, scope=None, client_id = None,
         auth_url = sp_oauth.get_authorize_url()
 
         # Working around webbrowser's printout
-        msg = "Opening %s in your browser" % auth_url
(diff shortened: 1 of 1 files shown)


adgud:
Why this hasn't been merged? it appears to be working on all platforms (I also tested it on Windows) while `subprocess.call(["open", auth_url])` does not.

SLongofono:
Has this project died?  Having the above fix would be much appreciated!

plamere:
thanks for the PR - the functionality has been incorporated via another PR, so this one has been closed.

PR #44, opened 2015-03-14
Python 3 compat and per-request oauth state

mbirtwell:
Fixes for python 3:
- Absolute imports in spotipy.**init**
- use six to find urlencode
- pass client_id/client_secret in post body instead of header during get_access_token.
  A bit of a cop-out I don't really understand why it wasn't working as a header.
  There was something strange happening with the unicode encoding I thought I fixed that
  but then I just kept getting invalid client errors back from spotify. Putting it in the body worked
  first time though

I've kind of mixed in with this making state a parameter of get_authorize_url and removing state and scope
  from get_access_token - I feel bad this should really be another commit.
  These parameters aren't required get_access_token according to the spotify docs and what do you know it works
  with out.
  I also think that it's quite useful if it's convenient to specify the state in the get_authorize_url function
  this is mostly because I'm using that to stash a next page url.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,2 @@
-requests==2.3.0
+requests>=2.3.0
+six>=1.8.0
--- a/spotipy/__init__.py
+++ b/spotipy/__init__.py
@@ -1,2 +1,2 @@
 VERSION='2.0.1'
-from client import Spotify, SpotifyException
+from spotipy.client import Spotify, SpotifyException
--- a/spotipy/oauth2.py
+++ b/spotipy/oauth2.py
@@ -58,8 +58,8 @@ class SpotifyOAuth(object):
                 if 'scope' not in token_info or self.scope != token_info['scope']:
                     return None
 
-                if self._is_token_expired(token_info):
-                    token_info = self._refresh_access_token(token_info['refresh_token'])
+                if self.is_token_expired(token_info):
+                    token_info = self.refresh_access_token(token_info['refresh_token'])
 
             except IOError:
                 pass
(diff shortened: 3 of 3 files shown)


dsimandl:
As a random developer who just tried to use this on Python 3.4, I completely support these changes.

plamere:
hey @mbirtwell - apologies for neglecting this PR for so long. I haven't made the jump to python 3 myself, so these kind of PRs have always been lower priority.  Since it has been so long, and there have been a few other python 3.0 passes made over the library, I'm not sure if this PR is still relevant.  If you still have interest and if you have a minute if you could let me know if you think this PR is indeed still worth incorporating.  Thanks.

mbirtwell:
I'd like to see it merged. I think it's perfectly good work. Andy python 3 is starting to gather pace. If you think you'd look at it I'll fix the merge conflicts and split up the commits to be a bit more logical.

mbirtwell:
Looks like at least some of the problems, which inspired me to make this PR have already been fixed by other people and merged. I tried running the unit tests under python 3 and had a couple of problems. One was fixed by updating the requests library. Although I was already using one newer than is required in your requirements.txt. The other problem appears to be a simple failure to update the unit tests when a function name was changed. I've opened PR #144 to fix that. 

The unit tests now run clean on python 3.4. I'll test it for my use-case later this week.

Although I would like to quickly point out that the current python 2/3 compatibility code is all using try: ... except ImportError and if sys.version tests. It might be possible to make it a bit cleaner using six.

mbirtwell:
I've split this up PR up in to 3 #147 and #148, which address the minor tweaks to the library which previously I'd rashly included with my python 3 changes. And #149, which simplifies the python 2/3 compatibility code as I suggested above. 

I recognise you might not want to introduce a dependency on six, but it's there if you want it.

PR #54, opened 2015-06-04
Support for getting tracks from albums with more than 50 tracks

jsundram:
Albums with greater than 50 tracks (e.g.  spotify:album:1vAbqAeuJVWNAe7UR00bdM) don't have all their tracks returned from the `album_tracks` API. I added offset/limit params to the api client to support these albums, as well as test to ensure that the client and underlying API works.

Apologies for some unrelated whitespace changes; my editor automatically strips whitespace from the ends of lines.


shillshocked:
How can we know if there are more than 50 results using the offset?

PR #58, opened 2015-06-22
Update client.py

bnlucas:
`.next()` and `.previous()` wouldn't walk through results properly using something like:

``` python
sp = spotipy.Spotify()

def get_tracks(artist, locale=None):
    result = sp.search(q=artist, type='album', limit=10, market=locale)

    while True:

        # do something with the results.

        result = sp.next(result['albums'])

        if result is None:
            break
```

Both methods were update to go off the `offset`, `limit`, and `total`, of `result` instead of the `next` url returned.

`market` was also added as an argument in `.search` as the API allows to search in a specific market.

### DIFF
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -7,6 +7,7 @@ import base64
 import requests
 import json
 import time
+import re
 
 ''' A simple and thin Python library for the Spotify Web API
 '''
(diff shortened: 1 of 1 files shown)


plamere:
I believe that the bug in the API has been fixed, so this change is no longer needed.

PR #65, opened 2015-09-24
Use 'Retry-After' header if present.

mrben:
When the Web API responds with a status code of `429` it sets the `Retry-After` header to let you know how long to wait before retrying. https://developer.spotify.com/web-api/user-guide/#rate-limiting

This PR honours that header and defaults back to the original `delay` variable if not present.

[requests headers](http://docs.python-requests.org/en/latest/api/#requests.Response.headers) use a case insensitive dictionary so we don't have to worry about the case that Spotify respond with.

### DIFF
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -12,10 +12,15 @@ import time
 '''
 
 class SpotifyException(Exception):
-    def __init__(self, http_status, code, msg):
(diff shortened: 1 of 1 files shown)


mrben:
Have rebased this against master so no longer conflicts.

plamere:
Thanks for the PR!

PR #62, opened 2015-08-05
Add 'market' parameter to the search endpoint

mattiamaestrini:
Fixes #57.

Added the missing `market` parameter to the `search` endpoint as described in the [Search for an Item](https://developer.spotify.com/web-api/search-item/) documentation.

PR #74, opened 2015-12-27
Add market option to limit on the availability in a certain market

dveeden:
I was searching for an album which had 3 results with the same artist and album name, but only one was available for the correct market.


brki:
I'd also like to have this option.

This pull request seems to be a duplicate of #62, no?

dveeden:
Yes indeed a duplicate of #62

PR #79, opened 2016-01-17
make version match history

jsundram:
version history says we're at 2.3.7; `VERSION` told a different story; make them match.


plamere:
catching up on these PRs.

PR #88, opened 2016-04-08
include refresh_access_token method to the library

eugenio412:
in this way it's possible to refresh token even if it's not cached but saved somewhere else


eugenio412:
patch

PR #96, opened 2016-06-01
Support endpoint for modifying playlist details

nathancoleman:
I added support for the API endpoint [here](https://developer.spotify.com/web-api/change-playlist-details/) which allows for the modification of playlist details such as the name and whether the playlist is marked as public and/or collaborative.

I also did some general cleanup. In a few cases, this included modifying args defaulted to mutable values (mostly `arg=[]`) to avoid the pitfalls described [here](https://pythonconquerstheuniverse.wordpress.com/2012/02/15/mutable-default-arguments/).

Please review: @plamere

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -302,6 +302,7 @@ Spotipy authored by Paul Lamere (plamere) with contributions by:
   - Steve Winton // swinton
   - Tim Balzer // timbalzer 
   - corycorycory // corycorycory 
+  - Nathan Coleman // nathancoleman
 
 License
 =======
--- a/examples/change_playlist_details.py
+++ b/examples/change_playlist_details.py
@@ -1,0 +1,38 @@
+
+# Modify the details of a playlist (name, public, collaborative)
+
+import sys
+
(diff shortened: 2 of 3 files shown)


nathancoleman:
@plamere I needed this change in my project [here](https://github.com/nathancoleman/feedback). Would you mind taking a look at this PR?

nathancoleman:
@plamere bump

plamere:
Hey Nathan ... this is looking good. I'd be extra pleased if you could include an example in the examples dir that show how someone could use the feature, and extra points too if you added a test.

carlosedp:
Any news on this PR? I'm switching to @nathancoleman branch while this isn't merged.

nathancoleman:
@carlosedp I just need to add some tests and an example use case. I should be able to get that done today.

nathancoleman:
@plamere I've added an example use case. A test for this endpoint would require someone maintaining a playlist strictly for running the test and then having the test run as the playlist owner, so I have not included a test for this change. Please review.

PR #98, opened 2016-06-09
Added market parameter to user_playlist_tracks

brodin:
According to https://developer.spotify.com/web-api/get-playlists-tracks/ this endpoints supports an market option. This is important to specify if you want to ensure 'preview_url' on your tracks (https://github.com/spotify/web-api/issues/148)


brodin:
Thanks @plamere, but when will this be published? :)

PR #111, opened 2016-08-18
starting browser for spotify authentication

OliverDiedrich:
Replaces `subprocess.call(["open", auth_url])` which only works on OS X with `webbrowser.open(auth_url, new=1)` which also works on Linux and Windows.


plamere:
thanks for the PR. The webbrowser technique has been incorporated via another PR.

Issue #112, opened 2016-08-23
Spotify Exception - Audio Features


patalanov:
I've been successufully retrieving audio_features from Spotify's `recommendations endpoint`, like so:

`features = sp.audio_features(tracks_ids)`

However, if I provide a list of id's retrieved using `search endpoint` and pass them in the exact same fashion to `sp.audio_features()`, I get this:

`spotipy.client.SpotifyException: http status: 414, code:-1 - https://api.spotify.com/v1/audio-features?ids=2ujuUDGDg6t5zsN6WZ3CFp,2EOThFm0IhwLkdpQzbvunO,40f9IDTMDpFf3CnTcPhY5F,78qoim2GGUkspkxV8kUtVv,4WZJ5W4gZJXvlqgliLkPCn,4mHS76nxzGrwo25KpzQwiX,1jRHh8JcdUV9zTiAmyzhU9...etc`

is there a different authentication flow? 

I have been using:

```
client_credentials_manager = SpotifyClientCredentials('id', 'secret')
sp = spotipy.Spotify(client_credentials_manager=client_credentials_manager)
sp.trace=False
```

what is going on here?

Baschdl:
It's not only issue #113, there are more problems. I will try to fix it.

Baschdl:
Pull request #114 fixes this issue

PR #114, opened 2016-08-23
Fixed bug in fetching of audio features

Baschdl:
Now it's possible to fetch the audio features of only one song. Fixes issue #112


plamere:
this PR seems to be mislabeled, it is not just fixing an audio features issue, but is also adding proxy support

PR #115, opened 2016-08-23
Added support for proxy

Baschdl:
Added support for proxies over `requests`

### DIFF
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -45,7 +45,7 @@ class Spotify(object):
     max_get_retries = 10
 
     def __init__(self, auth=None, requests_session=True,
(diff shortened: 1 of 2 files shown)


plamere:
I'm reluctant to merge this PR. I don't have a  sense for how widespread the need of proxy support is. I also don't have any way to test the changes to make sure that they actually do what they are supposed to do.  If you can make a case for why proxy support is important and help me figure out a good way to test the changes, that will help me decide whether proxy support is worth the effort.

Baschdl:
Everyone behind a company firewall with proxy needs this, otherwise spotipy won't work. You could setup a proxy in your local environment, I don't know a better way to test it.

PR #121, opened 2016-10-18
add support /me/playlists method

squallcs12:
A shortcut for API at https://developer.spotify.com/web-api/get-a-list-of-current-users-playlists/

So that we can get the current user playlist without require to know current user info

### DIFF
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -331,7 +331,7 @@ class Spotify(object):
         '''
         return self._get('users/' + user)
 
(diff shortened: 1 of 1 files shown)


spotipy/client.py
```diff
@@ -330,6 +331,14 @@ def user(self, user):
         '''
         return self._get('users/' + user)
 
+    def me_playlists(self, limit=50, offset=0):
```
plamere: I'm going to change this to 'current_user_playlists' to match other endpoints that work on the current user.

PR #122, opened 2016-10-18
Add market parameter to tracks() call

jotsif:
Get several tracks endpoint supports market parameter which allows getting linked tracks. https://developer.spotify.com/web-api/get-several-tracks/


plamere:
thanks for the PR!

PR #124, opened 2016-10-24
Recommendations seeds parameters formatted as a comma seperated list

steinitzu:
Hello Paul
Thanks for making Spotipy :)

I noticed when using multiple seed artists of different genres to fetch recommendations, that the resulting tracks were all sort of samey.   
Turns out that Spotify API does not support multiple URL parameters in the form of "seed_artists=xxxx&seed_artists=yyyy&...", this results in Spotify API ignoring everything but the seed.   

I wrapped the 'seed_x' parameters in an ','.join() call to fix the problem.

### DIFF
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -144,7 +144,7 @@ class Spotify(object):
                         delay += 1
                 else:
                     raise
-            except Exception as e: 
+            except Exception as e:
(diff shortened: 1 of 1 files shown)


plamere:
merged steinitzu's pr

PR #145, opened 2017-01-04
Update function to make it compatible to Python 3.

carlosedp:
The class `basestring` does not exist in Python 3. Change to `str`.


plamere:
won't this fail in python2 if name is a unicode string?

carlosedp:
I think this will match all cases.

plamere:
thanks, this looks like it will work.

PR #148, opened 2017-01-07
Make is_token_expired public

mbirtwell:
This method is useful for checking whether tokens that have been stored in cookies etc have timed out. So it's useful for library users to have access to it.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -352,6 +352,7 @@ Spotipy authored by Paul Lamere (plamere) with contributions by:
   - Tim Balzer // timbalzer 
   - corycorycory // corycorycory 
   - Nathan Coleman // nathancoleman
+  - Michael Birtwell // mbirtwell
 
 License
 =======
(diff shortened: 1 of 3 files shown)


plamere:
for consistency, we should make SpotifyClientCredentials._is_token_expired()

plamere:
Our tests for Spotipy are still quite sparse - it will help as we accelerate the PRs if we add tests for new features. Of course, some of these features especially around oauth are hard to write tests for with out creating mocks, which I'm not a fan of.  But if you do add tests when adding new functionality, that'd be helpful. Also, make sure you run all the tests before submitting PRs ;)

plamere:
Also, feel free to add yourself to the Contribute section in index.rst

mbirtwell:
I can make that function public too, that seems fine. Now that you point it out, i notice that it's implementation is identical. We could make a free function to centralise that, removing the currently implementation from the classes would change the API in an unhelpful way though.

As to running the tests. I ran the ones that run through tox before pushing my PRs and ran the authtests.py file in python 2/3. I didn't run the other test files (I have now and they all pass). I also ignored a test failure in authtests, which didn't seem relevant to my changes and also fails on master. test_user_playlist_follow, which perhaps only passes when run as you it seems to make the current spotify user follow a playlist and then checks that you are following the playlist.

PR #149, opened 2017-01-07
Use six to simplify python 2/3 compatibilty code where possible

mbirtwell:
This doesn't actually add any new python 2/3 compatibility. What it does do is use the six module where ever possible to simplify the code that needs to be 2/3 aware.


plamere:
I'm reluctant to add a dependency but this seems worth it.

PR #156, opened 2017-01-16
Add support for public playlists

cmcnutt1:
Allow the ability to get information from a public playlist. In order to do this, you must have the User ID of the playlist owner, and the playlist ID (both of these can be found in the playlist URL). 

This is achieved through the added public_playlist function. Usage example, where sp is the spotipy.Spotify object :

```
playlist = sp.public_playlist(EXAMPLE_USER_ID, EXAMPLE_PLAYLIST_ID)
playlistTracks = playlist['tracks']['items']
for item in playlistTracks:
    print(item['track']['id'])
```

Example above will print the track ID for every song in that playlist

### DIFF
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -675,6 +675,18 @@ class Spotify(object):
         r = self._put('me/albums?ids=' + ','.join(alist))
         return r
 
+    def public_playlist(self, user, playlist_id):
+        ''' Return public playlist, given the owner user ID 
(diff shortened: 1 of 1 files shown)


plamere:
I think the method 'user_playlists' already has this functionality

plamere:
See this example

https://github.com/plamere/spotipy/blob/master/examples/read_a_playlist.py

PR #169, opened 2017-03-08
Add cache_path as an argument to prompt_for_user_token()

### LINKED ISSUE #167: Add cache_path as a parameter to prompt_for_user_token()
Currently `prompt_for_user_token()` sets `cache_path` to `.cache-username` under the user's current directory. This is problematic when using `prompt_for_user_token()` in other scripts since it will be looking for/creating `.cache-username` files in whatever directory the user happens to be in. This causes the user to have to re-authenticate and have duplicate files all over.

`prompt_for_user_token()` should allow the caller to set the `cache_path` or possibly even default to creating it inside the user's `$HOME`.

I'm happy to submit a PR for these if either seem like a good idea.

jessejoe:
Fixes #167 

This allows overriding the default relative location of `cache_path` when calling `prompt_for_user_token()`. External scripts will be able to avoid creating a `.cache-username` file in whatever directory the user happens to be in when the script is ran.

I personally think `cache_path` should default to `os.path.join(os.path.expanduser('~'), '.cache-' + username)` so `cache_path` always references the same file in the user's `$HOME`, but that would probably break a lot of existing users.

### DIFF
--- a/spotipy/util.py
+++ b/spotipy/util.py
@@ -8,7 +8,7 @@ import spotipy
 import webbrowser
 
 def prompt_for_user_token(username, scope=None, client_id = None,
-        client_secret = None, redirect_uri = None):
+        client_secret = None, redirect_uri = None, cache_path = None):
     ''' prompts the user to login if necessary and returns
         the user token suitable for use with the spotipy.Spotify 
         constructor
@@ -20,6 +20,7 @@ def prompt_for_user_token(username, scope=None, client_id = None,
          - client_id - the client id of your app
          - client_secret - the client secret of your app
          - redirect_uri - the redirect URI of your app
+         - cache_path - path to location to save tokens
 
     '''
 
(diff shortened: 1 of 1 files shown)


KevinMGranger:
Ideally, the cache path would be in the `$XDG_CACHE_DIR`, which defaults to `~/.cache/`, so something like `~/.cache/spotipy-$USERNAME`

jessejoe:
@KevinMGranger I don't think that's cross-patform (i.e. Windows). Could always adopt the `pip` method though: https://github.com/dstufft/pip/blob/master/docs/user_guide.rst#config-file

`pip` is just using `appdirs` from https://github.com/ActiveState/appdirs, see https://github.com/dstufft/pip/blob/master/pip/utils/appdirs.py#L2

eriktaubeneck:
It would be great to merge this, and open a new issue to discuss/decide on a sane default. This PR doesn't introduce breaking changes and adds flexibility (and unblocks a blocker for me.)

plamere:
sorry for the 6 month delay, finally getting to look at my spotipy backlog. Looks like a good PR, thanks!

Issue #170, opened 2017-03-12
"webbrowser" cannot work on headless devices


shantanugoel:
There was a dependency added on "webbrowser" module with #139. However, this makes spotipy unusable on headless installs as there is no browser present there and even the package doesn't exist in pip for such devices and spotipy fails to even run.

The earlier method of subprocess call, in comparison, was a minor inconvenience on systems where browser didn't open automatically as url could be clicked/copied from the terminal. I request to revert this change, if possible.

shantanugoel:
Closing as this is fixed in #173

PR #173, opened 2017-03-21
Fix execution on headless systems where webbrowser module may not be available

shantanugoel:
On several headless systems (including but not limited to optware, entware packages for NASs, routers, etc), python builds do not include webbrowser module to save space since there's no browser on these systems. 
As mentioned in #170, currently spotipy does not run on these systems because it fails in the import statement. This fix will allow spotipy to run well on headless systems as well.
I've tested it to work on my linux PC as well as on my Asus routers (optware and entware) and Synology NAS (entware and synology python).

### DIFF
--- a/spotipy/util.py
+++ b/spotipy/util.py
@@ -5,7 +5,6 @@ from __future__ import print_function
 import os
 from . import oauth2
 import spotipy
-import webbrowser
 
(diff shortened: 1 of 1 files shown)


plamere:
thanks for the PR

Issue #174, opened 2017-03-26
Add support for the Recently Played endpoint


liammac:
Spotify has added a recently played endpoint. It would be great if support for this could be added

https://developer.spotify.com/web-api/web-api-personalization-endpoints/get-recently-played/

gwynnebaer:
Seems like a dupe. Can we close it? #175 resolves this.

PR #175, opened 2017-03-27
added recently_played

liammac:
Added recently played endpoint https://github.com/plamere/spotipy/issues/174


dddanmar:
Fixes my problem, cheers.

For context, maybe change `recently_played` to `user_recently_played`?

plamere:
there was another implementation of this in a recent PR that better matches the method naming convention, so I went with that one, but thanks much for the PR!

PR #179, opened 2017-04-01
Token cache handles null scopes correctly

richardmitic:
Handle null scopes gracefully by creating empty sets where necessary and comparing them as before.


rbdixon:
+1... hit this very bug.

PR #181, opened 2017-04-04
Add currently playing endpoint

mdbraber:


roman-ku:
@plamere Can this be merged?

plamere:
thanks for the PR, there are a few PRs that implement this endpoint. The one by lorenzofar seems to match the method naming scheme best, so I am going with that one.

PR #182, opened 2017-04-07
Add connect api endpoint

happyleavesaoc:
Support for the new Spotify Connect API: https://developer.spotify.com/web-api/web-api-connect-endpoint-reference/

### DIFF
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -995,7 +995,10 @@ class Spotify(object):
                 - device_id - device id to append
         '''
         if device_id:
-            path += "&device_id=%s" % device_id
+            if '?' in path:
+                path += "&device_id=%s" % device_id
+            else:
+                path += "?device_id=%s" % device_id
         return path
 
     def _get_id(self, type, id):
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -842,6 +842,162 @@ class Spotify(object):
         id = self._get_id('track', id)
         return self._get('audio-analysis/'+id)
 
+    def devices(self):
+        ''' Get a list of user's available devices.
+        '''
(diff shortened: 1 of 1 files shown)


tgurr:
As discussed in https://github.com/home-assistant/home-assistant/issues/7286 would it be possible to get this feature integerated and a new spotipy version released? This would ease adoption for distributions. Personally I can say I've tested spotipy (in combination with home assistant and different hardware, namely an Amazon FireTV and a Google Nexus 4 with the official Android Spotify app) with this pull request and it works really well so I'd love to see this feature getting merged.

durzo:
I had to make a small fix to this PR in order to get it working.

_append_device_id adds "&device_id=%s" onto the path but this breaks if there are no existing query strings (missing "?"). causing spotify API to return 404

This fixed for me:
```
    def _append_device_id(self, path, device_id):
        ''' Append device ID to API path.

            Parameters:
                - device_id - device id to append
        '''
        if device_id:
            if '?' in path:
                path += "&device_id=%s" % device_id
            else:
                path += "?device=id=%s" % device_id
        return path
```

jhaucke:
For those who wonder why changing the current device not working correctly:
With "?device_id=%s" instead of "?device=id=%s" it will work even better ;)

happyleavesaoc:
@durzo @jhaucke updated PR with that fix.

lenisko:
@plamere could you hit accept ;-) ?

dirtshell:
Will accept tommorow. Very drunk right now. And can't figure things out right now.

plamere:
Nice PR, thanks!

happyleavesaoc:
Thanks for the merge!

PR #183, opened 2017-04-11
Added support for following an artist or user

moyiz:
Must receive a list (consistent with the behavior of other calls)


plamere:
thanks for the pr

PR #184, opened 2017-04-11
Fix quote formatting according to PEP 257

Acrobot:
PEP 257 says to use three " signs instead of ' signs.


plamere:
neato, didn't know tht those had to be double-quotes, thanks for the pr

PR #185, opened 2017-04-14
Changes critical is not for !=

ncordon:
Hi, I was having trouble deploying an application in AppEngine using `spotipy` Client Credentials Flow. After a lot of debugging, I've realized there's two critical comparisons to a response status 200 made with `is not`, when they should be made with `!=` since they're comparisons against a number.

Regards


plamere:
nice catch, thanks for the PR

PR #189, opened 2017-05-09
Adding user current playback and available devices endpoint

danielgondim:


happyleavesaoc:
Note that https://github.com/plamere/spotipy/pull/182 covers this and more.

PR #196, opened 2017-06-03
Playlist description support

nathancoleman:
I added support for the description field which was newly added to the [create playlist](https://developer.spotify.com/web-api/create-playlist/) and [update playlist details](https://developer.spotify.com/web-api/change-playlist-details/#disqus_thread) endpoints. I also modified the related examples to include this new field.

A test for this endpoint would require someone maintaining a playlist strictly for running the test and then having the test run as the playlist owner, so I have not included a test for this change.

Please review: @plamere


AnsgarSchmidt:
Great! I was just going to add it. Hope this pull request will be processed soon. Thanks for the effort!

PR #199, opened 2017-06-27
add capability to get devices, play a track through a given device

asetapen:
...and some cleanup


asetapen:
Sorry - i targeted the wrong remote.

PR #204, opened 2017-07-22
Passing credentials into method

waterafternoon:
clear example of using credentials without having to use environment variables


plamere:
thanks for the pr!

Issue #232, opened 2017-11-14
Documentation typo


Hurstwood:
There is a typo in the documents at http://spotipy.readthedocs.io/en/latest/#authorization-code-flow

`util.prompt_for_user_token(username,scope,client_id='your-app-redirect-url',client_secret='your-app-redirect-url',redirect_uri='your-app-redirect-url')`

Should be...

`util.prompt_for_user_token(username,scope,client_id='your-app-client-id',client_secret='your-app-client-secret',redirect_uri='your-app-redirect-url')`

Or to be clearer to new programmers you could have...

```
clientID = 'Paste your client ID here'
clientSecret = 'Paste your client secret here'
redirectURL = 'Paste your redirect url here'  # We/Spotify suggest http://localhost:8888/callback/ or http://localhost/

token = util.prompt_for_user_token(username,scope,client_id='your-app-client-id',client_secret='your-app-client-secret',redirect_uri='your-app-redirect-url')
```

stephanebruckert:
Doc is now fixed

PR #234, opened 2017-11-16
Fix documentation typo

alexbotello:
Fixes typo specified in #232


jayeve:
The `client-id` parameter string is also wrong. #239

PR #243, opened 2018-01-04
util.py: add local http server to serve redirects

csernazs:
When spotify redirects the browser to localhost, serve the page with a
simple http server implemented by HTTPServer. By this way, we can obtain
the tokens the oauth service want so provide us, and no need to input the
user the redirect URL.

This method only works when the application in spotify developer dashboard
is configured to redirect the requests to localhost.

### DIFF
--- a/spotipy/oauth2.py
+++ b/spotipy/oauth2.py
@@ -17,11 +17,13 @@ import time
 import warnings
 
 import requests
-from spotipy.util import CLIENT_CREDS_ENV_VARS
+from spotipy.util import CLIENT_CREDS_ENV_VARS, get_host_port
 
 # Workaround to support both python 2 & 3
 import six
 import six.moves.urllib.parse as urllibparse
+from six.moves.BaseHTTPServer import HTTPServer, BaseHTTPRequestHandler
+from six.moves.urllib_parse import urlparse, parse_qsl
 
 
 class SpotifyOauthError(Exception):
@@ -296,6 +298,42 @@ class SpotifyOAuth(SpotifyAuthBase):
     def _make_authorization_headers(self):
         return _make_authorization_headers(self.client_id, self.client_secret)
 
+    def _open_auth_url(self):
+        auth_url = self.get_authorize_url()
+
+        try:
+            import webbrowser
+            webbrowser.open(auth_url)
+            print("Opened %s in your browser" % auth_url)
+        except webbrowser.Error:
+            print("Please navigate here: %s" % auth_url)
+
+    def _get_auth_response_interactive(self):
+        self._open_auth_url()
+        print("")
+        print("")
+
+        try:
+            response = raw_input("Enter the URL you were redirected to: ")
+        except NameError:
+            response = input("Enter the URL you were redirected to: ")
+        print("")
+        print("")
+        return self.parse_response_code(response)
+
+    def _get_auth_response_local_server(self, redirect_port):
+
+        with start_local_http_server(redirect_port) as server:
+            self._open_auth_url()
+            server.handle_request()
+
+        if server.auth_code is not None:
+            return server.auth_code
+        elif server.error is not None:
+            raise SpotifyOauthError("Received error from OAuth server: {}".format(server.error))
+        else:
+            raise SpotifyOauthError("Server listening on localhost has not been accessed")
+
     def get_auth_response(self):
         print(
             """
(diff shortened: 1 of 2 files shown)


jykong:
I just tested this out, and it works on my local machine. I definitely think this is a more intuitive authentication flow and should be the default. Early on when I was just getting started using spotipy, I got hung up on figuring out that I had to copy paste a non-working URL into the input field in the console.

@csernazs @plamere What are your thoughts for making the this the default method?

csernazs:
Well, I like it for my project (of course), but enabling it as default and deprecating the other one has some downside also:
* the redirect URL must be registered in spotify developer dashboard and this should point to a non-privilegized port (eg. above 1024) on localhost. I'm not sure whether this is in practice, I have my own application registered in spotify so I can specify here any localhost URL I can use but if I want to make my script work everywhere, it can fail when the port is already taken by other process, so we cannot bind.
* not every case the redirect URL is set to localhost, eg if I want to use someone's other app registration for this, it may have a different URL and this case only the copy&paste would work IMO.

There is also a "mixed" approach here:
- if the redirect URL Is set to localhost:port then we can try to bind to it and if that succeeds, we can proceed with this one
- otherwise, fall back to the copy&paste way (possibly with a message sent to the user to try to set the redirect URL to localhost if he/she is the owner of the app in spotify)

jykong:
Alternatively, we can add another parameter to the function that let's the user specify whether or not to use the localhost server (i.e. bindHttpServer=False). With the default value set to False, it would maintain backwards compatibility while allowing others to opt-in by setting it to True. For new users, the examples in the docs could be updated include bindHttpServer=True and explain how the port to bind to is parsed from the redirect URI.

hillrnate:
Hi @jykong and @csernazs , I would like to use this approach. In order to do so, is it just a matter of calling `util.obtain_token_localhost` instead of `util.prompt_for_user_token`?

Harrison97:
Hey! I am hosting a new master branch for this repo since the old creator has been MIA for 2 years. Hoping to get this repo back on it's feet and in good standing. I've been working on updating tests and adding all new functionality and have merged some of the PR's that I thought were sufficient for merging. If your PR didn't make it, resubmit it please.

Thanks!

https://github.com/Harrison97/spotipy

stephanebruckert:
Hi @csernazs, thanks for that. 2 years later, the package is maintained again. I know that's a late answer but is there a chance you can give this a new look and fix conflicts?

stephanebruckert:
Just saw you are updating this! Thanks a lot. What do you think of these 2 duplicates? 
 - https://github.com/plamere/spotipy/pull/301
 - https://github.com/plamere/spotipy/pull/342

When merging yours, will we be able to close these ones? Is there anything good to take from them?

csernazs:
Hi!

I think both #301 #342 are virtually the same as mine, they provide the same functionality. #301 seems to provide some protection in multi-threaded environment which is clearly missing from mine.

I would like to leave it to you the decision to pick the one which is the best for you. Alternatively you can create a 4th one by merging the best parts you think from the three.

I used this library to move from google play music to spotify so I have no hard dependencies on it.

Also, one think I noticed: maybe you want to drop python 2 support, struggling with python 2 unittests, and six were a necessity in the past but definitely not in the present and future.

csernazs:
Speaking of what we can take from the other two: I think the difference is so subtle that these are virtually the same. I don't really understand the locking stuff in #301, if you see any usefulness in that, I can merge it into mine.
Also, the texting is very subjective, so if you think that there's something we can improve in my case, let me know.

stephanebruckert:
Sorry this is my fault, I skim read your PR and I just saw now that you are adding new methods and not updating old methods. I would agree with @jykong that this could be made the default method. The idea being that someone who updates spotipy, would not have to update their code to "enjoy" this feature. I would definitely go for making it the default or go for the "mixed approach" you mentioned, but in all cases it should be available to everyone with none or little code change (i.e. one extra parameter max when calling oauth).

Let me know if you have the time to change it, otherwise no problem we might just fall back to #301 or #342. Sorry again and in all cases thank you for responding present with today's changes

csernazs:
Let me see what can I do. I think that what I said in my comment before is doable (https://github.com/plamere/spotipy/pull/243#issuecomment-356399300). The user will need to specify the port where the server is listening and also which is configured in the dashboard, so we can fall back to the old behavior when it is not specified (eg. it can be backward compatible).

csernazs:
I submitted a draft code for this, it works fine for me, however there are still some points to address:
* tests are missing :)
* prompt_for_user_token would need to be renamed to something else, but that would break the API

jykong:
Hey all,
Happy to see you all are maintaining this. It's been too long for me to
remember the details of the changes you are discussing, and I don't have
the bandwidth right now to refresh myself. I trust you're smart people and
will make sensible choices!

stephanebruckert:
I've ran the tests from your repo and I could see it work as expected: instead of prompting for the authentication URL it just moved on to the next tests automatically 👍 

I've closed https://github.com/plamere/spotipy/pull/301 and https://github.com/plamere/spotipy/pull/342 and am going to merge your PR now! I might add a commit for the changelog and a bit of JS to automatically close the browser window.

Thanks a lot for your work @csernazs, this is great!

csernazs:
:champagne: thank you!

PR #346, opened 2019-01-28
Use `urllib`'s `quote` to encode user ids in urls #307

### LINKED ISSUE #307: user id's are not URL percent encoded
user id's are not URL encoded and results in some 404 errors when usernames contain symbols. Can be solved simply by using `urllib.parser.quote` but it seems this repository is no longer updated so I am posting this more as a public service announcement than anything else.

loisaidasam:
Fixes #307

### DIFF
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -7,6 +7,14 @@ import requests
 import json
 import time
 
+try:
+    # Python 3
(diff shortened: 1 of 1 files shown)


loisaidasam:
Closing in favor of continued https://github.com/plamere/spotipy/pull/350

PR #372, opened 2019-04-27
changed refresh token timer to ensure it refreshes on time

rotywonka:
I've noticed that whenever I use refresh the access tokens, it doesn't refresh it on time. By refreshing it every 10 minutes, it leaves no room for error of having the access token 'expired'.  

Signed-off-by: Tommy Tan <tommyytan@gmail.com>

### DIFF
--- a/spotipy/oauth2.py
+++ b/spotipy/oauth2.py
@@ -23,7 +23,7 @@ def _make_authorization_headers(client_id, client_secret):
 
 def is_token_expired(token_info):
     now = int(time.time())
-    return token_info['expires_at'] - now < 60
+    return token_info['expires_at'] - now < 1200
 
 
 class SpotifyClientCredentials(object):
@@ -67,7 +67,7 @@ class SpotifyClientCredentials(object):
 
     def _request_access_token(self):
         """Gets client credentials access token """
-        payload = { 'grant_type': 'client_credentials'}
+        payload = {'grant_type': 'client_credentials'}
 
         headers = _make_authorization_headers(self.client_id, self.client_secret)
 


Harrison97:
Hey! I am hosting a new master branch for this repo since the old creator has been MIA for 2 years. Hoping to get this repo back on it's feet and in good standing. I've been working on updating tests and adding all new functionality and have merged some of the PR's that I thought were sufficient for merging. If your PR didn't make it, resubmit it please.

Thanks!

https://github.com/Harrison97/spotipy

MaZderMind:
I don't think this does what you think it does.
The access tokens expiry is only currently checked, whenever `spotipy.util.prompt_for_user_token` is called. An expired token is then refreshed. This does however only work, when the application does not run longer then the remaining lifetime of the token, when `spotipy.util.prompt_for_user_token` is called.

When the Application runs longer then the remaining lifetime (or longer then 3600s / 1h), the token will expire no matter how tight the window in `is_token_expired` is.

I think the actual problem you tried to solve is actually solved by #428
I know this PR is quite old, so I don't really expect you to check that, but my guess is that this PR is obsoleted by #428

stephanebruckert:
I'm reopening this, just realised it's not a bad idea. For example it would prevent this situation to happen:

 1. program runs and use token1
 2. 59 minutes, program runs again and re-use token1
 3. 2 minutes later, the program is still running and a call to spotify fail

Another possibility is to only refresh the token when a "token outdated" exception is raised, also avoiding any unneeded calls to /refresh

MaZderMind:
@stephanebruckert that is not what will happen.
1. 0:00 Token is created (1h lifetime)
2. 0:50 App Terminates
3. 0:55 App starts again
5. 0:56 API-Call is being made. App identifies the Token is still good and uses the Token. API-Call succeeds.
6. 1:05 API-Call is being made. App identifies the Token is >59 Minutes old and requests a new one. API-Call succeeds.

stephanebruckert:
@MaZderMind agreed, I said that the scenario would be "prevented"

stephanebruckert:
This is not needed anymore thanks to https://github.com/plamere/spotipy/pull/435, closing

PR #393, opened 2019-08-17
Add two missing api endpoints relating to playlists

germaindudek:
Two endpoints to get playist info and track. Difference between this one
and other playlist endpoints is that they don't require you to provide
a user_id


stephanebruckert:
Thanks a lot for your PR @germaindudek but I ended up improving https://github.com/plamere/spotipy/pull/422 which is only slightly different that your version

Issue #323, opened 2018-10-02
`Spotify.artist_albums` acts as `Spotify.artist` when passing HTTP artist URL


ritiek:
When `Spotify.artist_albums` is given an HTTP artist URL like `https://open.spotify.com/artist/7oPftvlwr6VrsViSDV7fJY?si=M3PrzRC4TBOZu8YyLYc-tA`, it behaves same as `Spotify.artist` and instead of returning the albums, it just returns artist info. However, it works fine if artist URI is passed like `spotify:artist:7oPftvlwr6VrsViSDV7fJY` and then it returns albums as expected.

For example:

```python
>>> import spotipy
>>> import spotipy.oauth2 as oauth2

>>> credentials = oauth2.SpotifyClientCredentials(
...         client_id=CLIENT_ID,
...         client_secret=CLIENT_SECRET)

>>> token = credentials.get_access_token()
>>> spotify = spotipy.Spotify(auth=token)

>>> greenday_http = 'https://open.spotify.com/artist/7oPftvlwr6VrsViSDV7fJY?si=M3PrzRC4TBOZu8YyLYc-tA'
>>> artist_as_http = spotify.artist(greenday_http)

>>> greenday_uri = 'spotify:artist:7oPftvlwr6VrsViSDV7fJY'
>>> artist_as_uri = spotify.artist(greenday_uri)

>>> artist_as_http == artist_as_uri
True

>>> artist_albums_as_http = spotify.artist_albums(greenday_http)
# `Spotify.artist_albums` acts same as `Spotify.artist`
>>> artist_albums_as_http == artist_as_http
True

>>> artist_albums_as_uri = spotify.artist_albums(greenday_uri)
# `Spotify.artist_albums` does not act same as `Spotify.artist` as expected
>>> artist_albums_as_uri == artist_as_http
False
```

ritiek:
Fixed by #420.

PR #420, opened 2020-01-13
Trim '?<key>' when extracting ID from HTTP URLs

### LINKED ISSUE #365: Inconsistent return object of `user_playlist_tracks`
If given a playlist_id `user_playlist_tracks` returns a [paging object](https://developer.spotify.com/documentation/web-api/reference/object-model/#paging-object),
if given a full url it returns [playlist object (full)](https://developer.spotify.com/documentation/web-api/reference/object-model/#playlist-object-full). Tested on [this](https://open.spotify.com/user/didrik0510/playlist/7ehrAeEI4icxsqlvGX02gW?si=LaDb4nMATFqmeGVgvhHObQ) playlist.

Code snipper below demonstrates the difference:
```python
import spotipy
from spotipy import util

token = util.prompt_for_user_token(
    username,
    client_id=_client_id,
    client_secret=_client_secret,
    redirect_uri=_redirect_uri,
)
sp = spotipy.Spotify(auth=token)

url = "https://open.spotify.com/user/didrik0510/playlist/7ehrAeEI4icxsqlvGX02gW?si=LaDb4nMATFqmeGVgvhHObQ"
id_ = "7ehrAeEI4icxsqlvGX02gW"
print(sp.user_playlist_tracks(username, url).keys())
print(sp.user_playlist_tracks(username, id_).keys())
```

Having it written in the documentation would be nice, with the explanation what input results in what output type.

ritiek:
If this excess part isn't trimmed from the ID, it can cause unexpected
results when calling some API methods.

For example; this code now correctly returns all albums for an artist:
```
artist = 'https://open.spotify.com/artist/7oPftvlwr6VrsViSDV7fJY?si=M3PrzRC4TBOZu8YyLYc-tA'
artist_albums = sp.artist_albums(artist)
```

Previously, this code mimicked same behaviour as calling
`spotify.artist`, but now will return albums as expected
with this commit.

Fixes #365 and #323.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+ - Fixed inconsistent behaviour with some API methods when
+   a full HTTP URL is passed.
+
 ### Changed
  - Fixed invalid calls to logging warn method
  - `mock` no longer needed for install. Only used in `tox`.
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -1128,7 +1128,7 @@ class Spotify(object):
             if type != itype:
                 self._warn('expected id of type %s but found type %s %s' %
                            (type, itype, id))
-            return fields[-1]
+            return fields[-1].split('?')[0]
         return id
 
     def _get_uri(self, type, id):


ritiek:
Should be ready to merge if everything looks ok!

stephanebruckert:
My bad, just double checked and it’s not the same

PR #428, opened 2020-01-21
Automatic refresh of Authorization Code Flow Tokens in long-running Applications

### LINKED ISSUE #87: Support refreshing token
What is the recommendation of this team for how people should refresh tokens? As it stands all examples are written as though you've just gotten your tokens from a prompt. If I go deep into oauth2.py I can find a private method to refresh tokens however, that requires a whole different object that the standard spotipy instance does not need.

This is a critical part of the library. How do you anticipate this flow working?

### LINKED ISSUE #263: Refreshing a token with authorization code flow
I have a long-running script using spotipy. After an hour (per the Spotify API), my access token expires. I am catching this successfully, but I don't know where to go from there in regards to actually refreshing the token. I am using the authorization code flow, not client credentials. Here's how I authorize:

`token = util.prompt_for_user_token(username,scope=scopes,client_id=client_id,client_secret=client_secret, redirect_uri=redirect_uri)

sp = spotipy.Spotify(auth=token)`

All refresh examples I've seen involve an `oauth2` object (ex. `oauth.refresh_access_token()`), and the docs list only that function as a method of refreshing your token. It's my understanding that with authorization code flow, you don't need an `oauth` object (because you authenticate with `prompt_for_user_token()`. If that's the case, how do I refresh my token?

MaZderMind:
This PR implements automatic refreshing of Authorization Code Flow Tokens.
Currently an `access_token` requested via the Authorization Code Flow have a lifetime of 3600 seconds (1 hour) after which a new `access_token` needs to be requested using the 'refresh_token`.

The Logic in `util.prompt_for_user_token` looks for a matching set of tokens in the `cache_path` and if none is found, it requests both an `access_token` and a `refresh_token` from Spotify and stores them in the `cache_path`. If one is found, the `access_token` is checked for validity and if it is expired, a new one is requested and updated in the `cache_path`.

This process works fine for short running applications, because the Token is automatically refreshed, whenever `util.prompt_for_user_token` is invoked. For long running applications (with a runtime greater then the remaining lifetime of the `access_token` upon startup) this poses a problem, because the `access_token` is never refreshed, once `util.prompt_for_user_token` has returned.

To refresh the token later, the initialized instance of `oauth2.SpotifyOAuth` is required. This instance is created in `util.prompt_for_user_token` but destroyed when the method returns.

This PR changes the return-value of `util.prompt_for_user_token` to return the initialized `oauth2.SpotifyOAuth` instead of the `access_token`. This instance is accepted as the `auth`-argument to the `spotipy.Spotify` constructor, so code that was just passing the token-string don't have to be changed.

Furthermore `oauth2.SpotifyOAuth` is changed to also cache the `access_token` in-memory, so that accessing it via the oauth2-instance does not have to hit the filesystem.

The method `Spotify._auth_headers` is responsible for adding the Auth-Token to the HTTP Request-Headers. It is changed to actively request a valid `access_token` from the oauth2 instance. The `oauth2.SpotifyOAuth` class is changed to check the validity of the Token before returning and, if required, requesting a new Token.

An Example of a long-running process which before exposed the problem is added. It does not run for multiple hours, refreshing the Token when it has expired before issuing any other request.

This probably fixes #263 and fixes #87
It probably also closes #428 by obsoleting it

### DIFF
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -63,7 +63,7 @@ class Spotify(object):
         """
         Creates a Spotify API client.
 
-        :param auth: An authorization token (optional)
+        :param auth: An authenticated OAuth2 Instance
         :param requests_session:
             A Requests session object or a truthy value to create one.
             A falsy value disables sessions.
@@ -94,7 +94,8 @@ class Spotify(object):
 
     def _auth_headers(self):
         if self._auth:
-            return {'Authorization': 'Bearer {0}'.format(self._auth)}
+            token_info = self._auth.get_cached_token()
+            return {'Authorization': 'Bearer {0}'.format(token_info['access_token'])}
         elif self.client_credentials_manager:
             token = self.client_credentials_manager.get_access_token()
             return {'Authorization': 'Bearer {0}'.format(token)}
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -95,7 +95,8 @@ class Spotify(object):
     def _auth_headers(self):
         if self._auth:
             token_info = self._auth.get_cached_token()
-            return {'Authorization': 'Bearer {0}'.format(token_info['access_token'])}
+            access_token = token_info['access_token']
+            return {'Authorization': 'Bearer {0}'.format(access_token)}
         elif self.client_credentials_manager:
             token = self.client_credentials_manager.get_access_token()
             return {'Authorization': 'Bearer {0}'.format(token)}
--- a/spotipy/oauth2.py
+++ b/spotipy/oauth2.py
@@ -16,7 +16,6 @@ import sys
 import time
 
 import requests
-
 # Workaround to support both python 2 & 3
 import six
 import six.moves.urllib.parse as urllibparse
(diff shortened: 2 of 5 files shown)


MaZderMind:
I'll look into the failing tests.

spotipy/oauth2.py
```diff
@@ -213,7 +222,7 @@ def parse_response_code(self, url):
     def _make_authorization_headers(self):
         return _make_authorization_headers(self.client_id, self.client_secret)
 
-    def get_access_token(self, code):
+    def request_access_token(self, code):
```
stephanebruckert: This looks like a public method so if possible I would avoid renaming it

MaZderMind: Yes, you're right about that. I intended to add another `get_access_token` method which would just return the possibly refreshed token, but found that `get_cached_token` is basically doing this exact thing. I removed this change from the commit and force-pushed an update.

MaZderMind:
I removed the offending method-name change from the PR and added a Line to the changelog. I also extracted a duplicate code-fragment into a new `_refresh_token_if_expired` method in `oauth2.SpotifyOAuth`

stephanebruckert:
Thank you!

ritiek:
@MaZderMind Sorry, but this had to be reverted since it seems causing #430. We could re-merge this once the issue is fixed.

spotipy/client.py
```diff
@@ -94,7 +94,9 @@ def __init__(self, auth=None, requests_session=True,
 
     def _auth_headers(self):
         if self._auth:
-            return {'Authorization': 'Bearer {0}'.format(self._auth)}
+            token_info = self._auth.get_cached_token()
+            access_token = token_info['access_token']
+            return {'Authorization': 'Bearer {0}'.format(access_token)}
```
ritiek: These changes seem to cause #430. Is there a good reason we need these changes?

spotipy/client.py
```diff
@@ -63,7 +63,7 @@ def __init__(self, auth=None, requests_session=True,
         """
         Creates a Spotify API client.
 
-        :param auth: An authorization token (optional)
+        :param auth: An authenticated OAuth2 Instance
```
stephanebruckert: For backward-compatibility, this parameter should still be "An authorization token"

spotipy/util.py
```diff
@@ -103,9 +103,7 @@ def prompt_for_user_token(username, scope=None, client_id=None,
...
         code = sp_oauth.parse_response_code(response)
-        token_info = sp_oauth.get_access_token(code)
+        sp_oauth.get_access_token(code)
+
     # Auth'ed API request
-    if token_info:
-        return token_info['access_token']
-    else:
-        return None
+    return sp_oauth
```
stephanebruckert: Also here, to be backward-compatible, we should still return an access token

stephanebruckert:
I'm working on this in a new branch https://github.com/plamere/spotipy/tree/change/auto-refresh-token

PR #431, opened 2020-01-23
Adding playlist_cover - playlist_upload_cover_image's style fixed

CyanBook:


spotipy/client.py
```diff
@@ -422,42 +426,6 @@ def playlist_upload_cover_image(self,
                          payload=image_b64,
                          content_type="image/jpeg")
 
-    def user_playlist(self, user, playlist_id=None,
```
stephanebruckert: We want to keep the old endpoints because they still work and people might still be using them

spotipy/client.py
```diff
@@ -405,6 +405,10 @@ def playlist_tracks(self, playlist_id, fields=None,
         return self._get("playlists/%s/tracks" % (plid),
                          limit=limit, offset=offset, fields=fields,
                          market=market)
+    
+    def playlist_cover(self, playlist_id=None):
```
stephanebruckert: A comment under the method (see method below) will help to automatically generate documentation for that method on https://spotipy.readthedocs.io/

spotipy/client.py
CyanBook: They can avoid to update to this version or update their code, using old functions isn't a good idea

spotipy/client.py
CyanBook: edited

spotipy/client.py
stephanebruckert: People should still be able to upgrade, be able to use new features (performance, security, etc), and migrate from old functions to new functions when they find the time to do it. For now, these two endpoints have deprecation warnings

stephanebruckert:
Merged, thank you @CyanBook

Issue #321, opened 2018-09-29
Cache file default naming convention


zsiegel92:
For some reason I'm unable to submit a pull request for this repository. I suggest changing line 49 of `util.py` from:

```python
cache_path = cache_path or ".cache-" + username
```
to:
```python
cache_path = cache_path or ".cache-" + username + "-" + str(scope)
```

Several issues have been opened regarding the need to interact with the browser more than once to refresh tokens when using `util.py`. I believe it is because  the token cache file is overwritten when tokens of different scopes are created using the default token cache file naming in `util.py`. Of course, users  of this library can manually enter a `cache_path`, but I believe other users ran into confusion, as I did. I don't believe the Spotipy documentation says enough about this function to justify making the default behavior of `util.py` nonfunctional for anyone creating multiple tokens.

Thank you for making this library. It is a useful tool for getting started interacting with the Spotify API quickly.

JikkyHD:
I support that suggestion very much. 
Also, you are not able to `util.prompt_for_user_token()` again after doing that once, because it will fail with 
`Traceback (most recent call last):

    token = util.prompt_for_user_token(username, scope)

/spotipy/util.py", line 56, in prompt_for_user_token

    token_info = sp_oauth.get_cached_token()

/spotipy/oauth2.py", line 135, in get_cached_token

    if 'scope' not in token_info or not self._is_scope_subset(self.scope, token_info['scope']):

/spotipy/oauth2.py", line 156, in _is_scope_subset

    needle_scope = set(needle_scope.split())

AttributeError: 'NoneType' object has no attribute 'split'`

sanchitcop19:
God bless you @zsiegel92

stephanebruckert:
Closing because too many unanswered questions https://github.com/plamere/spotipy/pull/432#issuecomment-578504990

PR #432, opened 2020-01-24
Save auth cache for multiple scopes

truenite:
Saves cache in different files for different scopes

### DIFF
--- a/spotipy/util.py
+++ b/spotipy/util.py
@@ -64,7 +64,10 @@ def prompt_for_user_token(username, scope=None, client_id=None,
         ''')
         raise spotipy.SpotifyException(550, -1, 'no credentials set')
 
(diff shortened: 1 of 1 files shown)


stephanebruckert:
I like this idea (https://github.com/plamere/spotipy/issues/321) but I have some questions:

 - does this break backward-compatibility? Let's say, my app has been running for 1 year refreshing the same token, will upgrading spotipy with this change require me to sign in again? Is there a possibility to make that transition smoothly?
 - what happens if scopes are identical but not ordered in the same way?
 - what about file size and the path+filename 260 character limit on some systems (windows)?
 - why not keep the last token as "the valid one"? which could be what a user is expecting from an app: to keep only the last authorized one

stephanebruckert:
Going to close this in favor of "only keeping the last token is the way to go", any objection?

Issue #287, opened 2018-05-10
Best way to use Spotify inside an API?


Jeoffreybauvin:
Hi,

I'm using Spotipy in a Flask API (full rest).  I'm trying to use util.prompt_for_user_token for login, and it's working.

But now, I want to create an entrypoint /authenticate, which send me back the URL that  util.prompt_for_user_token is trying to open in my browser.

I don't see any information in the documentation to do so. Can I catch this URL without opening my browser ?

MaZderMind:
@Jeoffreybauvin In the Case described by you I would not use `util.prompt_for_user_token`, because it clearly states to be a utility to get you up and running quickly. When Implementing a custom RESt-API, I would instead operate the required Client-Classes directly.

When removing all the input-checking and comments from `util.prompt_for_user_token` there is not a lot left:
```
def redirect_user_to_oauth(username, scope=None, client_id=None,
                          client_secret=None, redirect_uri=None,
                          cache_path=None):

…
    sp_oauth = oauth2.SpotifyOAuth(client_id, client_secret, redirect_uri,
                                   scope=scope, cache_path=cache_path)
…
    token_info = sp_oauth.get_cached_token()
    if not token_info:
        auth_url = sp_oauth.get_authorize_url()
        return auth_url


def handle_callback(code):
    token_info = sp_oauth.get_access_token(code)
    return token_info['access_token']
```

As an API-Implementor you have to provide the requisite options to `oauth2.SpotifyOAuth` and have it request an `auth_url` for you. Depending on the Implementation of your REST-API I would return this URL to my client in order for it to open it in a popup or otherwise have my user access it.

After the OAuth2 dance, the user is redirected to the `redirect_uri` you provided, which should be some url in your application (it might also be a local server if your writing a native app). The supplied code is then transferred to your REST-API in a second call, which can use it to request `token_info` on you users behalf.

stephanebruckert:
For now this provides good inspiration to solve this https://github.com/plamere/spotipy/pull/435#issuecomment-583890341

stephanebruckert:
Just added an example here https://github.com/plamere/spotipy/blob/master/examples/app.py

andrewhansen89:
> Just added an example here https://github.com/plamere/spotipy/blob/master/examples/app.py

Doesn't this example need a cache_path or username?

PR #435, opened 2020-02-03
Auto-refresh AuthCode flow token.

stefanondisponibile:
I loved @stephanebruckert 's idea of [having similar handlers for the two different authorization flows](https://github.com/plamere/spotipy/issues/263#issuecomment-578539713).

I think it would be even better having one one unique `auth_manager` inside the `Spotify` client, and infer it from the parameters. This also would allow a smooth and backward compatible pass to this.

What do you think?

### DIFF
--- a/examples/simple4.py
+++ b/examples/simple4.py
@@ -2,6 +2,7 @@ import spotipy
 import os
 from pprint import pprint
 
+
 def main():
     auth_manager = spotipy.SpotifyOAuth(
         client_id=os.getenv("SPOTIPY_CLIENT_ID"),
@@ -12,6 +13,7 @@ def main():
     spotify = spotipy.Spotify(auth_manager=auth_manager)
     me = spotify.me()
     pprint(me)
+
 
 if __name__ == "__main__":
     main()
--- a/examples/simple4.py
+++ b/examples/simple4.py
@@ -1,0 +1,17 @@
+import spotipy
+import os
+from pprint import pprint
+
+def main():
+    auth_manager = spotipy.SpotifyOAuth(
+        client_id=os.getenv("SPOTIPY_CLIENT_ID"),
+        client_secret=os.getenv("SPOTIPY_CLIENT_SECRET"),
+        redirect_uri=os.getenv("SPOTIPY_REDIRECT_URI"),
+        username=os.getenv("SPOTIPY_CLIENT_USERNAME"),
+    )
+    spotify = spotipy.Spotify(auth_manager=auth_manager)
+    me = spotify.me()
+    pprint(me)
+
+if __name__ == "__main__":
+    main()
--- a/examples/simple4.py
+++ b/examples/simple4.py
@@ -2,15 +2,8 @@ import spotipy
 import os
 from pprint import pprint
 
-
 def main():
-    auth_manager = spotipy.SpotifyOAuth(
-        client_id=os.getenv("SPOTIPY_CLIENT_ID"),
-        client_secret=os.getenv("SPOTIPY_CLIENT_SECRET"),
-        redirect_uri=os.getenv("SPOTIPY_REDIRECT_URI"),
-        username=os.getenv("SPOTIPY_CLIENT_USERNAME"),
-    )
-    spotify = spotipy.Spotify(auth_manager=auth_manager)
+    spotify = spotipy.Spotify(auth_manager=spotipy.SpotifyOAuth())
     me = spotify.me()
     pprint(me)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  - Support for `playlist_cover_image`
  - Support `after` and `before` parameter in `current_user_recently_played`
  - CI for unit tests
+ - Automatic `token` refresh
+ - `auth_manager` and `oauth_manager` optional parameters added to `Spotify`'s init.
+ - Optional `username` parameter to be passed to SpotifyOAuth, to infer a `cache_path` automatically
+ - Optional `as_dict` parameter to control `SpotifyOAuth`'s `get_access_token` output type. However, this is going to be deprecated in the future, and the method will always return a token string
+
+### Changed
+  - Both `SpotifyClientCredentials` and `SpotifyOAuth` inherit from a common `SpotifyAuthBase` which handles common parameters and logics.
 
 ## [2.7.1] - 2020-01-20
 
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -78,8 +78,8 @@ class Spotify(object):
             SpotifyClientCredentials object
         :param oauth_manager:
             SpotifyOAuth object
-        :param auth_manger:
-            
+        :param auth_manager:
+            SpotifyOauth object or SpotifyClientCredentials object
         :param proxies:
             Definition of proxies (optional)
         :param requests_timeout:
@@ -113,7 +113,9 @@ class Spotify(object):
         if auth_manager is not None:
             self._auth_manager = auth_manager
         else:
-            self._auth_manager = self.client_credentials_manager or self.oauth_manager
+            self._auth_manager = (
+                self.client_credentials_manager or self.oauth_manager
+            )
 
     def _auth_headers(self):
         if self._auth:
@@ -165,7 +167,10 @@ class Spotify(object):
                 except BaseException:
                     msg = "error"
                 raise SpotifyException(
-                    r.status_code, -1, "%s:\n %s" % (r.url, msg), headers=r.headers
+                    r.status_code,
+                    -1,
+                    "%s:\n %s" % (r.url, msg),
+                    headers=r.headers,
                 )
 
             try:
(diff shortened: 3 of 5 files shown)


stephanebruckert:
I will review and respond later, but for now to fix the lint you can just do:

    pip install autopep8
    autopep8 --in-place --aggressive --recursive .
    pip install flake8
    flake8 .

It's missing from the README

examples/simple4.py
```diff
@@ -0,0 +1,19 @@
...
+import os
+from pprint import pprint
+
+
+def main():
+    auth_manager = spotipy.SpotifyOAuth(
+        client_id=os.getenv("SPOTIPY_CLIENT_ID"),
+        client_secret=os.getenv("SPOTIPY_CLIENT_SECRET"),
+        redirect_uri=os.getenv("SPOTIPY_REDIRECT_URI"),
+        username=os.getenv("SPOTIPY_CLIENT_USERNAME"),
```
stephanebruckert: If you get some free time, it would be amazing to make those optional (like for Client Credentials), so we could just do:

    spotipy.Spotify(auth_manager=spotipy.SpotifyOAuth())

ritiek:
Looks cool! I'll hopefully be able to review by evening.

examples/simple4.py
stefanondisponibile: Yep, I had the idea too! I'll make this soon 👍

stefanondisponibile:
> This is looking great! I'm still trying to think of a scenario where being forced to be prompted for the token could be an issue. Could it be a problem when trying to use spotipy in an API? [#287 (comment)](https://github.com/plamere/spotipy/issues/287#issuecomment-576896586). In this case we would need something like `prompt=True`
> 
> What do you think @ritiek?

@stephanebruckert  in my opinion the best way to handle the authorization flow independently is using an authorization manager independently too, and just use the `Spotify` client by passing the `auth` token, or by initializing a manager once you've retrieved it programmatically for the first time.

stephanebruckert:
@stefanondisponibile the problem is that some people might already be doing this https://github.com/plamere/spotipy/issues/287#issuecomment-576896586, so if possible we would be looking for a solution that does not break that, do you see what I mean?

stefanondisponibile:
Hey @ritiek, do you hav any updates on this?

ritiek:
Oops, sorry. It seems like review comments need to be submitted twice?

Anyway, this looks good to me. Seems to automatically refresh token nicely as well. I've left a few trivial review comments.

stefanondisponibile:
Cool! I pushed the updates, can't wait to have this merged 🥳

stephanebruckert:
Hey, sorry I was ill for most of the week, unable to do much. I just had a better look at how SpotifyOAuth works, I really want to merge the PR but there is still something backward-incompatible in it.

@stefanondisponibile:
> in my opinion the best way to handle the authorization flow independently is using an authorization manager independently too, and just use the Spotify client by passing the auth token, or by initializing a manager once you've retrieved it programmatically for the first time.

Please confirm if I understand correctly. You would make the previous `SpotifyOAuth` (no prompt) a new class, something like `SpotifyOAuthNoPrompt` so we can use prompt in the updated `SpotifyOAuth` from this PR? Unfortunately the priority is to keep the previous implementation working because some people might already be using it on a server where there is no way to "prompt a user" and that a minor update would block them.

So I see two possible solutions to fix this PR:

 - use `SpotifyOAuth(prompt=True)` as initially suggested,
 - add a `SpotifyOAuthPrompt` auth manager (not `SpotifyOAuthNoPrompt`)

Let me know if we are both understanding it correctly?

Thanks again for the work you put in this PR, can't wait to see it merged!

stefanondisponibile:
Hey @stephanebruckert , good to have you back, sorry you was ill!
I'm not sure I explained my point properly (I wouldn't create more classes). Thinking of passing a `prompt` parameter is an option, but probably I'm missing part of the problem.

For now, the main compatibility issue after the changes I made, is that `SpotifyOAuth.get_access_token` is now returning a `str` instead of a `dict`. This might be an issue if someone is using this method by himself.

If instead, the problem is on the prompting part, could you please try to explain that to me again?
What I don't understand is this.
**SCENARIO 1**:
Let's say I've made a script that does something with spotipy, let's say something like `examples/simple4.py`:

```python
import spotipy
import os
from pprint import pprint


def main():
    spotify = spotipy.Spotify(auth_manager=spotipy.SpotifyOAuth())
    me = spotify.me()
    pprint(me)


if __name__ == "__main__":
    main()
```

The first time I run this script it will tell me to login to spotify, gave my permission for the scopes, and will send me back to a url that I have to paste in the console to get the `code` that is passed in the current `SpotifyOAuth.get_access_token` method. This is cool because I don't have to this myself on the script, the spotipy package will "drive" me through that.
The second time I run the script the token's been cached, so it will work without asking anything until the end of time, since the token gets refreshed automatically too.

** SCENARIO 2 **:
Let's say I've got the exact same script as above. But this time we're on a remote server, there's no user in front of the screen to paste things. So yeah, we don't want to prompt, but the first time I ever run the script I **need** to get an access token, how do I do that? I need to call `SpotifyOAuth.get_access_token`, but for that I need a `code`. How to get that `code` is my responsibility: I will need to send the user who's interacting with the app to the authorization url, and catch back the code myself from the redirect uri. To do that I could also think of using the builtin `get_authorization_code`, or just parse it myself, or whatever. When I have that code I can finally pass it to the `get_access_token` method and forget about it forever.

So could you explain me again the issue please?

Thank you :)

stephanebruckert:
Hey @stefanondisponibile! I agree with scenario 1 👍 

In the scenario 2,  `get_access_token` is an object method, I don't think you call it this way:

    SpotifyOAuth.get_access_token()

instead it should be called on an object of that class:

    oauth = spotipy.SpotifyOAuth()  # prompt here, cannot avoid it
    oauth.get_access_token()

So even if you get the code independently, the prompt would open before. There must be ways to adapt the way it works, but we cannot change it because it might already be used (for example in something that would look like https://github.com/plamere/spotipy/issues/287#issuecomment-576896586)

Do you see what I mean? Could I be missing something?

stefanondisponibile:
Sorry @stephanebruckert , but I still don't really get the problem.
Of course, `get_access_token` is an _instance_ method.

---

> `oauth = spotipy.SpotifyOAuth()  # prompt here, cannot avoid it`

why can't avoid it? You shouldn't receive any prompts by doing this.

---
 In fact, one should actually reflect what's in [@MaZderMind  comment](https://github.com/plamere/spotipy/issues/287#issuecomment-576896586).

Try this and see if that's the expected behaviour:

```python
"""
Of course:
0. pip install spotipy (from this branch or from my fork)
1. pip install flask
2. Export needed env variables:
  export SPOTIPY_CLIENT_ID=yourclientid
  export SPOTIPY_CLIENT_SECRET=yourclientsecret
  export SPOTIPY_CLIENT_USERNAME=yourusername
3. Add "http://localhost:5000" to your app's redirect uris.
4. Copy-paste this into an "app.py" file.
5. Run "flask run" and go to http://localhost:5000

"""

from flask import Flask, request, jsonify
import spotipy

app = Flask(__name__)

auth_manager = spotipy.oauth2.SpotifyOAuth()
spotify = spotipy.Spotify(auth_manager=auth_manager)

@app.route("/")
def index():
    if request.args.get("code"):
        auth_manager.get_access_token(request.args["code"])
    if not auth_manager.get_cached_token():
        auth_url = auth_manager.get_authorize_url()
        return f'<h2><a href="{auth_url}">Authorize this application.</a></h2>'
    return spotify.me()
```

stephanebruckert:
Indeed, that's my bad! For some reason I was thinking that `SpotifyOAuth()` would prompt, but it doesn't which makes the whole thing look very concise now. Perfect & sorry about that!

--------
> For now, the main compatibility issue after the changes I made, is that SpotifyOAuth.get_access_token is now returning a str instead of a dict. This might be an issue if someone is using this method by himself.

About this change, I think it's a good idea initially because ClientCredentials.get_access_token is returning a string too. But it looks like the dict returned by `SpotifyOAuth.get_access_token` was used in some places, [here](https://github.com/sedir/festify/blob/4793d25b332e1f2a85b32ff6dc8094fe16f8cfa9/app.py#L62) and possibly [in more places](https://github.com/search?q=get_access_token+spotipy+SpotifyOAuth&type=Code). Is it a lot of work to use a dict again?

Also feel free to add a changelog and we should be good to merge! Cheers

stefanondisponibile:
No worries @stephanebruckert , always better overchecking than the opposite 😉

---

Good catch! On other's code we might break if we return a string instead of a dictionary!
However, I like the idea of keeping ClientCredentials and OAuth manager "similar", so how about this solution instead: let's return a dictionary, just like before, but add an `as_dict` parameter, that's going to be deprecated in the future. With this parameter, we start using `as_dict=False` in `spotipy`'s internal codebase, and we'll remove that in the future, maybe with major releases.

I'm pushing these changes, so you can have a look.

stephanebruckert:
@stefanondisponibile sounds perfect!

stephanebruckert:
Very nice, merging, thank you @stefanondisponibile

MaZderMind:
Thank you all for following up on this 👍
Great to see the better implementation win 👌

stephanebruckert:
For now this provides good inspiration to solve issue https://github.com/plamere/spotipy/pull/435#issuecomment-583890341

louhow:
Just wanted to say, I was tweaking an old project that was leveraging this client and auth_manager let me get rid of so much code. Thanks for the contribution!

PR #437, opened 2020-02-09
Linting of core subpackage + clean up imports

sr-murthy:
* Linting of core package
* Clean up imports

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -7,6 +7,7 @@ deps=
     py27: mock
 commands=python -m unittest discover -v tests
 [flake8]
+ignore=E501
 exclude=
     .git,
     dist,
--- a/tox.ini
+++ b/tox.ini
@@ -7,7 +7,7 @@ deps=
     py27: mock
 commands=python -m unittest discover -v tests
 [flake8]
-ignore=E501
+max-line-length = 99
 exclude=
     .git,
     dist,
--- a/tox.ini
+++ b/tox.ini
@@ -11,4 +11,5 @@ ignore=E501
 exclude=
     .git,
     dist,
-    docs
+    docs,
+    examples
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -12,10 +12,10 @@ __all__ = [
 import json
 import sys
 import time
+import warnings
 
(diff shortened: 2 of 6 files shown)


tox.ini
```diff
@@ -7,7 +7,9 @@ deps=
     py27: mock
 commands=python -m unittest discover -v tests
 [flake8]
+ignore=E501
```
stephanebruckert: Not sure we want to ignore lines too long? Perhaps we should change the max line length instead?

sr-murthy: OK, sure, what do you recommend?

sr-murthy: I just run `flake8`.

The changes also include reordering imports into blocks - first block containing standard lib. imports, second block containing external 3rd party imports, third block containing internal imports. Also, ensuring ascending order within blocks.

Nothing significant in terms of functionality, but it would be nice to keep imports clean and tidy.

stephanebruckert: Cool you are also using flake8, I suggest to add this setup.cfg file then? https://stackoverflow.com/a/42325590/1515819

stephanebruckert: About the reordering of the imports, I totally agree it's nicer, but there is good chances that the next developer will "forget" to order them the same way. Need to see if we can add a rule that checks this automatically.

stephanebruckert: Actually use a `.flake8` file instead of `setup.cfg`, this way it will be obvious that we are using flake8 http://flake8.pycqa.org/en/2.6.0/config.html#per-project

sr-murthy: According to the documentation flake8 can be configured via the tox ini, so I'll just add the line length instruction and update the PR. I hope that's OK.

sr-murthy: A tool that parses modules to check for proper import block structure would be quite useful - who knows, perhaps someone out there has already written one!

stephanebruckert: Yes looks like there is https://stackoverflow.com/a/58804315/1515819 but it involves a pre-commit as it's not available as a flake8 rule.

I'm gonna merge this PR for now. We can still add that later if we've got time

Thanks for that

PR #438, opened 2020-02-10
Prompt option to always show a dialog when giving permissions to Spotify

JohannesPertl:
The ```show_dialog``` field is added as an option to util.prompt_for_user_token().

This is necessary if you want to authenticate multiple users in a row


stephanebruckert:
Thanks for that @JohannesPertl, looks good!

PR #452, opened 2020-03-12
'add to queue' endpoint added

Peter-Schorn:
see https://developer.spotify.com/documentation/web-api/reference-beta/#endpoint-add-to-queue

This is my first pull request, so please let me know if I've done anything wrong.

PR #453, opened 2020-03-13
add_to_queue endpoint added

Peter-Schorn:


Peter-Schorn:
Did I do it right?

stephanebruckert:
Oh, you should have used the same PR https://github.com/plamere/spotipy/pull/452. But that's fine, I will merge both. Thanks a lot!

Peter-Schorn:
So is there anything else I need to do?

Issue #347, opened 2019-01-29
Retry handling not implemented for POST


navels:
Seems that the retry handling is only implemented for GET.  It should be probably be moved into `client._internal_call` instead so that it covers all the API calls.

stephanebruckert:
Related PR https://github.com/plamere/spotipy/pull/271

PR #271, opened 2018-04-03
Implement retry handling for POST

hmhrex:

PR #457, opened 2020-03-24
Retries For All Endpoints 

Beahmer89:
**Purpose of PR**
I was looking through open issues when I noticed issue [347](https://github.com/plamere/spotipy/issues/347). I noticed that there was a pull request up for this [here](https://github.com/plamere/spotipy/pull/271). The solution works with adding it to each method function, but it would be cleaner if moved into  `_internal_call`. Also the logic could be cleaned up as well, so I decided to make a PR.

This PR is to allow Retries for all method calls by utilizing `urllib3.Retry()` to allow the following:
- Setting a default list of status codes to retry on or list the user define a list
- Default/user defined number of retries to allow
- Default/user defined backoff value for how long the requests should wait before retrying.
  - `{backoff factor} * (2 ** ({number of total retries} - 1))` found in [urllib3 docs](https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html)

By doing this is, it cleans up the retry logic a lot that is already implemented in the code

**What Was Done**
-  Setup function in client to build the the new requests session with the urllib3.Retry()
- Restructured  `_internal_call`
  - Now catching specific exceptions not just generic ones
  - No longer using request call as context manager
  - Calling raise for error after the request call and then decoding the response to json() all in one try...except
- Updated by using appropriate unittest methods that make things more clear rather than just using assertTrue and assertFalse
  - this was only done in the client files
- Reorganized client tests to be grouped according to their Api defined [here](https://developer.spotify.com/documentation/web-api/reference-beta/#category-browse)
- Updated README to include how one could contribute to spotipy and what variables they need to have set for the tests to run.
- Added requirement dependency for nose since it wasnt already defined in requirements.txt 

**Other Things to Note**
- By default I disable retries for the session/adapter for read errors and set it to `False` like they do by default in the requests library [here](https://github.com/psf/requests/blob/master/requests/adapters.py#L117)

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,23 @@ Then, create a Spotify object and call methods:
     for idx, track in enumerate(results['tracks']['items']):
         print(idx, track['name'])
 
+## Contributing
+If you would like to contribute to spotipy follow these steps:
+### Export the needed environment variables
+```bash
+    export SPOTIPY_CLIENT_ID=client_id_here
+    export SPOTIPY_CLIENT_SECRET=client_secret_here
+    export SPOTIPY_CLIENT_USERNAME=client_username_here # This is actually an id not spotify display name
+    export SPOTIPY_REDIRECT_URIa=http://localhost/ # Make url is set in app you created to get your ID and SECRET
+```
+
+### Create virtual enevironment, install dependencies, run tests:
+```bash
+$ virtualenv --python=python3.7 env
+(env) $ pip install requirements.txt
+(env) $ nosetests -xvs tests
+```
+
 ## Reporting Issues
 
 If you have suggestions, bugs or other issues specific to this library, file them [here](https://github.com/plamere/spotipy/issues). Or just send me a pull request.
--- a/README.md
+++ b/README.md
@@ -35,23 +35,6 @@ Then, create a Spotify object and call methods:
     for idx, track in enumerate(results['tracks']['items']):
         print(idx, track['name'])
 
-## Contributing
-If you would like to contribute to spotipy follow these steps:
-### Export the needed environment variables
-```bash
-    export SPOTIPY_CLIENT_ID=client_id_here
-    export SPOTIPY_CLIENT_SECRET=client_secret_here
-    export SPOTIPY_CLIENT_USERNAME=client_username_here # This is actually an id not spotify display name
-    export SPOTIPY_REDIRECT_URIa=http://localhost/ # Make url is set in app you created to get your ID and SECRET
-```
-
-### Create virtual enevironment, install dependencies, run tests:
-```bash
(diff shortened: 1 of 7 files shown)


ritiek:
Yup! I'll review in a moment.

README.md
```diff
@@ -35,6 +35,23 @@ Then, create a Spotify object and call methods:
...
         print(idx, track['name'])
 
+## Contributing
+If you would like to contribute to spotipy follow these steps:
+### Export the needed environment variables
+```bash
+    export SPOTIPY_CLIENT_ID=client_id_here
+    export SPOTIPY_CLIENT_SECRET=client_secret_here
+    export SPOTIPY_CLIENT_USERNAME=client_username_here # This is actually an id not spotify display name
+    export SPOTIPY_REDIRECT_URIa=http://localhost/ # Make url is set in app you created to get your ID and SECRET
```
ritiek: This looks like a typo. Should be `SPOTIPY_REDIRECT_URI` instead of `SPOTIPY_REDIRECT_URIa` I believe.

requirements.txt
```diff
@@ -1,3 +1,4 @@
 mock==2.0.0
+nose==1.3.7
```
ritiek: Do we really need `nose` though? Built-in `unittest` works fine too as mentioned in
https://github.com/plamere/spotipy/blob/8b84300597be027e0f7d8e40b67d1df04250db57/tox.ini#L8

README.md
```diff
@@ -35,6 +35,23 @@ Then, create a Spotify object and call methods:
     for idx, track in enumerate(results['tracks']['items']):
         print(idx, track['name'])
 
+## Contributing
```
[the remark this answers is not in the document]
Beahmer89: This sounds totally reasonable and will add that!

requirements.txt
Beahmer89: I dont think`nose` is needed. I forgot to check the `tox.ini` file to see how tests were run. I just used it as it was what I was familiar with.

Will remove from `requirements.txt` and use `unittest` command for example in new CONTRIBUTING.md

ritiek:
I've left a few minor review comments but otherwise looks good to me. Thanks a lot @Beahmer89, really appreciate all the work that went into this!

Beahmer89:
Thanks for the review @stephanebruckert and @ritiek! I am glad you guys like the approach for the retries! I forgot to add a note in the changelog about what I did. Let me know if want me to put up a PR for that. Sorry for the miss.

stephanebruckert:
Thanks to you @Beahmer89 for the great contribution. I also forgot about the changelog. Don't worry about it I will add it when bumping the version. Still feel free to open a quick PR if you want to, I will merge it straight away!

Issue #84, opened 2016-03-16
Use logging instead of print


eriktaubeneck:
Print statements, even in exception handling, makes it impossible for those using your client to actually control the output while building their own application. I'd be happy to provide a PR if it would be appreciated, but don't want to sync time into it if not.

Tommatheussen:
Did you ever put in the effort needed for this?

PR #461, opened 2020-04-02
Print -> Logging 

Beahmer89:
**Purpose of PR**
This is related to [Issue 84](https://github.com/plamere/spotipy/issues/84). Currently spotipy is using print statements in order to provide users with information about what is happening in the code. It is much more effective to use the logging library in python so that users are able to customize the output and see the logging within their app thats using spotipy.

**What was Done**
- Replaced all print statements with correct logging level
  - Removed `trace` attribute as it is no longer needed
- In order to test this I modified a couple of the examples to use `logging` and `argparse`
  - Not entirely sure if you would like the approach so I only modified a part of the examples. Would like to put up another PR with making all the examples consistent
- Removed examples of errors and error handling. These are not needed in my opinion as there are tests that cover this.
- **BUG** - Noticed with the addition of automatically handling the redirect that I was getting a `socket.error: [Errno 13] Permission denied` (even when I changed my app to use default port 80) 
  - Saw in [stackoverflow](https://stackoverflow.com/questions/24001147/python-bind-socket-error-errno-13-permission-denied) that you can just run on a higher port number. 
  - changed default port number from 80 -> 8080
    - updated CONTRIBUTING.md


spotipy/client.py
```diff
@@ -178,24 +172,12 @@ def _internal_call(self, method, url, payload, params):
             if payload:
                 args["data"] = json.dumps(payload)
 
-        if self.trace_out:
```
stephanebruckert, line 181: We can keep these logs and print them as debug, e.g. `logger.debug`

spotipy/oauth2.py
```diff
@@ -25,6 +24,8 @@
 from six.moves.BaseHTTPServer import HTTPServer, BaseHTTPRequestHandler
 from six.moves.urllib_parse import urlparse, parse_qsl
 
+LOGGER = logging.getLogger(__name__)
```
stephanebruckert, line 27: I suggest to make it lowercase: `logger`

spotipy/client.py
```diff
@@ -178,24 +172,12 @@ def _internal_call(self, method, url, payload, params):
...
-
         try:
             response = self._session.request(
                 method, url, headers=headers, proxies=self.proxies,
                 timeout=self.requests_timeout, **args
             )
 
-            if self.trace:  # pragma: no cover
-                print()
-                print("Request headers:", headers)
```
stephanebruckert, line 192: same as above

stephanebruckert:
Thanks!  👍

PR #475, opened 2020-04-16
user_playlist_add_episodes method was added.

ilkayDevran:
Using Spotify APIs on developer console, adding podcast(show) episodes in a user playlist is possible. However, in this version of the repo has only adding tracks into the user playlist method. For my own use, I added a method for this just like already developed for tracks.

### DIFF
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -683,6 +683,25 @@ class Spotify(object):
             position=position,
         )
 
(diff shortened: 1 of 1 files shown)


stephanebruckert:
Closing for now, let's say that it's still ok to use `user_playlist_add_tracks` for episodes

Issue #474, opened 2020-04-16
Authentication Process is missing "Enter the URL you were redirected to:" part 


AnySomebody1:
Hi,
in the recent git version (as well as in the 2.11 release) the part of the authentication process is missing where one has to enter the URL Spotify redirected to and thus it seems not possible to authenticate the app anymore.

jonascj:
I think it is this issue https://github.com/plamere/spotipy/issues/471

Use `http://localhost:8080` or instead of just `http://localhost` as your redirect url (in both code/envvars and in spotify app settings).

AnySomebody1:
Indeed, it is. With a port it works just fine. I suggest to change the docs accordingly and to modify "The URL doesn’t need to be publicly accessible, so you can specify “http://localhost/”."

jonascj:
I've created PR #477 which deals with this issue.

Issue #471, opened 2020-04-14
Following Getting Started docs, can't authenticate with "Authorization Code Flow"


mfeif:
Hi There.

I'm following verbatim the docs [here](https://spotipy.readthedocs.io/en/latest/#authorization-code-flow). I have all the env variables set, and a premium account and all of that. I'm using localhost as per the instructions.

The browser can't connect to the redirected URL (which is http://localhost/?code=AQCBPehFt16_roFv...{long hash}) I've pasted the URL from the browser into the console as per the docs, and press return, and nothing happens. No error messages, no feedback, nothing.

Looking around the code, it seems like it is supposed to launch an httpd there, but I can't see it.

If I ctrl-C the script, this traceback results:
```python
Traceback (most recent call last):
  File "example.py", line 13, in <module>
    token = util.prompt_for_user_token(username, scope)
  File "/Users/mjf/Library/Caches/pypoetry/virtualenvs/pifi-GA63gEPW-py3.7/lib/python3.7/site-packages/spotipy/util.py", line 91, in prompt_for_user_token
    code = sp_oauth.get_auth_response()
  File "/Users/mjf/Library/Caches/pypoetry/virtualenvs/pifi-GA63gEPW-py3.7/lib/python3.7/site-packages/spotipy/oauth2.py", line 366, in get_auth_response
    return self._get_auth_response_local_server(redirect_port)
  File "/Users/mjf/Library/Caches/pypoetry/virtualenvs/pifi-GA63gEPW-py3.7/lib/python3.7/site-packages/spotipy/oauth2.py", line 346, in _get_auth_response_local_server
    server.handle_request()
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/socketserver.py", line 294, in handle_request
    ready = selector.select(timeout)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/selectors.py", line 415, in select
    fd_event_list = self._selector.poll(timeout)
KeyboardInterrupt
```

I tried using an unprivileged port (http://localhost:8001/) and I get an "INVALID_CLIENT: Invalid redirect URI" message from Spotify.

It doesn't seem like the script is running anything on localhost for the browser to connect to.

I'm on python 3.7.2 and spotipy is 2.11.1, running on macos.

What am I missing? Thanks!

stephanebruckert:
Hi @mfeif, thanks for the report.

You were on the right track as it looks like it works only when a port is specified. 
[...]

PR #477, opened 2020-04-18
Fix issue with hanging authorization process if no port is provided (and updated docs)

jonascj:
PR #243 added local http server to accept the auth redirect/callback,  eliminating the need to copy and paste a URL during authorization (very cool, I love it). 

But from my point of view, and issue #471 and #474 hints at the same thing, PR #243 broke applications / scripts which used redirect uri `http://localhost`, causing them to hang during authorization (never asking for the URL to be pasted/entered in the console/terminal).

Without this PR the fix is to change redirect uri to `http://localhost:8080` (specify a non-privileged port). This PR changes the authorization process logic slightly such that a local host redirect uri with port specified (e.g. `http://localhost:8080`) will start the local http server, but a local host uri *without* port specified (e.g. `http://localhost`) will fallback to the `Enter the URL you were redirected to:`-method.

With this PR old scripts/applications will continue to run and not hang unexpectedly. 

The problem arises because the current Spotipy release starts a local http server (to listen for the redirect/callback) as soon as it detects a redirect uri containing `localhost` or `127.0.0.1`. If no port is specified it defaults to port `8080` but since the request uri argument or envvar is still `http://localhost` Spotify redirects to `http://localhost:80/?code=...` and not `http://localhost:8080/?code=...`. So the local server never receives any requests and hence continues to block execution of the script.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -84,23 +84,19 @@ Or you can get the source from github at https://github.com/plamere/spotipy
 Getting Started
 ===============
 
-All methods require user authorization. You will need to
-register your app to get the credentials necessary to make authorized calls.
+All methods require user authorization. You will need to register your app at 
+`My Dashboard <https://developer.spotify.com/dashboard/applications>`_ 
+to get the credentials necessary to make authorized calls
+(a *client id* and *client secret*).
 
-Even if your script does not have an accessible URL you will need to specify one
-when registering your application which the Spotify authentication server will
-redirect to after successful login. The URL doesn't need to be publicly
-accessible, so you can specify "http://localhost/".
-
-Register your app at
-`My Applications
-<https://developer.spotify.com/my-applications/#!/applications>`_ and register the
-redirect URI mentioned in the above paragragh.
-
-*spotipy* supports two authorization flows:
+*Spotipy* supports two authorization flows:
 
   - The **Authorization Code flow** This method is suitable for long-running applications
     which the user logs into once. It provides an access token that can be refreshed.
+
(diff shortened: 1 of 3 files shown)


jonascj:
Another solution would be for Spotipy to silently (maybe with a warning) alter the redirect_uri argument or SPOTIPY_REDIRECT_URI by added a port to local host redirect URIs without port before starting the authorization process. I however like my approach better.

Also, I can see I have some test/check fails, let me look at those ...

jonascj:
@stephanebruckert Should I add/commit a changelog entry to this PR?

Should I document the changes under [Unreleased like so?

```
## [Unreleased]

### Changed

 - Updated the documentation to give more details on the authorization process and reflect 2020 Spotify Application jargon and practices.

 - The local webserver is only started for localhost redirect_uri which specify a port, i.e. it is started for `http://localhost:8080` or `http://127.0.0.1:8080`, not for `http://localhost`. 

### Fixed

 - Issue where using `http://localhost` as redirect_uri would cause the authorization process to hang.
```

jonascj:
@stephanebruckert  Would you consider going for a fast 2.11.2 release with this fix 2.11.1 breaks old scripts?

stephanebruckert:
Ok I can release after this is merged. Changelog looks good!

stephanebruckert:
Released in [2.11.2](https://github.com/plamere/spotipy/releases/tag/2.11.2)

jonascj:
Excellent, thanks for stepping up and maintaining the package!

PR #479, opened 2020-04-24
Added a method to update the auth token

rako77:
Auth token can't currently be updated without creating a new client. this change allows you to continue to use the same client object.


rako77:
Sorry about not running tests. trying to fix.

stephanebruckert:
Thanks for the PR!

PR #497, opened 2020-05-24
In '__init__' super() method should use static class name in SpotifyClientCredentials and SpotifyOAuth

esolitos:
Currently it is not possible to extend SpotifyClientCredentials and SpotifyOAuth without overrding completely the `__init__` method and avoid calling `super().__init__()`.

I am not an expert in Python so I could be completely wrong, however from my research this seem to be caused from the fact that in the classes' `__init__` they refer to the dynamic `self.__class__`, which when called by the extending class leads to to an infinite recursion.

This PR fixes the issue by setting the class name statically in the `super()` call.

### DIFF
--- a/spotipy/oauth2.py
+++ b/spotipy/oauth2.py
@@ -112,7 +112,7 @@ class SpotifyClientCredentials(SpotifyAuthBase):
         environment variables
         """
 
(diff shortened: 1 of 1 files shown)


stephanebruckert:
Thanks, merged!

PR #509, opened 2020-06-06
Check the state

foobuzz:
Hello,

The `SpotifyOAuth` class accepts an optional `state` parameter to be added to the authorization URL. However, this parameter wasn't checked when echoed back upon the redirection from Spotify, which pretty much made its usage useless. This PR simply adds the necessary check when `state` is used. I also refactored the extraction of the authentication code from the URL so that it's the same between interactive way and local server way.  

- Verify that the state received alongside the authorization code is consistent with the one that was sent
- Refactor URL parsing for the local server way and the interactive way
- Add tests for interactive way
- Allow `state` to be given to `util.prompt_for_user_token`

### DIFF
--- a/spotipy/oauth2.py
+++ b/spotipy/oauth2.py
@@ -322,11 +322,11 @@ class SpotifyOAuth(SpotifyAuthBase):
             Parameters:
                 - url - the response url
         """
-        url_split = url.split("?code=")
-        if len(url_split) <= 1:
+        _, code, _ = self.parse_oauth_response_url(url)
+        if code is None:
             return url
         else:
-            return url_split[1].split("&")[0]
+            return code
 
     @staticmethod
     def parse_oauth_response_url(url):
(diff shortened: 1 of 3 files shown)


spotipy/oauth2.py
```diff
@@ -316,17 +316,18 @@ def get_authorize_url(self, state=None):
 
         return "%s?%s" % (self.OAUTH_AUTHORIZE_URL, urlparams)
 
-    def parse_response_code(self, url):
```
stephanebruckert, line 319: public method that we must keep because others might be using it https://github.com/search?l=Python&q=%22parse_response_code%22+spotipy&type=Code

spotipy/oauth2.py
```diff
@@ -386,11 +389,6 @@ def get_auth_response(self):
                     'complete the authorization')
         return self._get_auth_response_interactive()
 
-    def get_authorization_code(self, response=None):
```
stephanebruckert, line 389: this is also a public method

foobuzz:
Hey, sorry about removing the public methods, that completely went over my head. I resurrected them!

stephanebruckert:
No worries @foobuzz, thanks a lot for the fixes, merging!

PR #516, opened 2020-06-19
enable market parameter for album_tracks

kenmacf:
enabling the optional market feature for get_album_tracks API:  https://developer.spotify.com/console/get-album-tracks

Adding in order to get the is_playable field for the track listing, which is not included if market is not set

### DIFF
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -339,7 +339,7 @@ class Spotify(object):
         trid = self._get_id("album", album_id)
         return self._get("albums/" + trid)
 
-    def album_tracks(self, album_id, limit=50, offset=0):
+    def album_tracks(self, album_id, limit=50, offset=0, market=None):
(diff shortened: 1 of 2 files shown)


spotipy/client.py
```diff
@@ -339,7 +339,7 @@ def album(self, album_id):
         trid = self._get_id("album", album_id)
         return self._get("albums/" + trid)
 
-    def album_tracks(self, album_id, limit=50, offset=0):
+    def album_tracks(self, album_id, limit=50, offset=0, market=None):
         """ Get Spotify catalog information about an album's tracks
 
             Parameters:
```
stephanebruckert, line 345: can update this with the new parameter

kenmacf: I've updated the album_tracks docstring--if there's another change needed, please clarify

stephanebruckert:
Merged, thank you

PR #518, opened 2020-06-25
Get all non-local tracks from a spotify playlist

Omglolyes:


stephanebruckert:
That's nice, maybe just rename the filename to `examples/playlist_all_non_local_tracks.py`?

examples/playlist_all_tracks.py
```diff
@@ -0,0 +1,23 @@
...
+tracks.extend(result['items'])
+
+# if playlist is larger than 100 song, continue loading it until end
+while result['next']:
+	result = sp.next(result)
+	tracks.extend(result['items'])
+
+# remove all local songs
+for item in tracks:
+	if item["is_local"] == True:
```
stephanebruckert, line 20: no need for `== True`

examples/playlist_all_tracks.py
```diff
@@ -0,0 +1,23 @@
...
+		tracks.remove(item)
+		
+return tracks
```
stephanebruckert, line 23: not sure this is useful, do you want `pp(tracks)` instead?

examples/playlist_all_tracks.py
```diff
@@ -0,0 +1,23 @@
+# get all non-local tracks of a playlist
+from spotipy.oauth2 import SpotifyClientCredentials
+import spotipy
+from pprint import pprint as pp
+
+sp = spotipy.Spotify(client_credentials_manager=SpotifyClientCredentials())
```
stephanebruckert, line 6: afaik removing tracks from playlists needs user authentication, did you try this code?

examples/playlist_all_tracks.py
Omglolyes: it `if item["is_local"]` should be enough?

stephanebruckert: Yes

Omglolyes:
i don't know if printing the result really gives any insight into what it does. especially pprint throws way to much information around. maybe only print track name and if it is local or not?

stephanebruckert:
You could print a count of the total local tracks deleted? 

> just deleted __ local track(s) in playlist ____

With the playlist you chose it might be 0, which is fine as people will at least have an example and will be able to replace the playlist id with their own

Issue #525, opened 2020-06-30
Search through all markets at once


stephanebruckert:
**Is your feature request related to a problem? Please describe.**

 - currently searches default to the US market and need separate requests for other markets (CA, GB, FR, DE...) 
 - is an issue when building playlists for anyone in the world
 - https://github.com/spotify/web-api/issues/194

**Describe the solution you'd like**
Add options to the `search` function to allow searching multiple markets at once:

 - allow passing an array of country codes 
 - allow passing "ALL"
 - have an option to return as soon as `n` results are found

**Additional context**
This will not be performant due to the amount of requests, appropriate warnings should be given

NLeRoy917:
I can give this a shot - I agree this might make a lot of requests, but in my experience, Spotify is pretty generous with API rate limiting

shillshocked:
How do we use this experimental functionality to search through all markets at once? I tried:
sp.search(search_str, offset=off, limit=50, type='album', market='ALL')

And got:
HTTP Error for GET to https://api.spotify.com/v1/search returned 400 due to Invalid market code

stephanebruckert:
@shillshocked https://spotipy.readthedocs.io/en/2.16.0/#spotipy.client.Spotify.search_markets

shillshocked:
OK, I try sp.search_markets(search_str, offset=off, limit=50, type='album') and get the same error. Also same results with:
sp.search_markets(search_str, offset=off, limit=50, type='album', market='ALL')

shillshocked:
I noticed that there are only 60 markets in client.py, but Spotify actually has at least 92 country codes. It should be updated to include them all, or you can just specify markets=['AL','DZ','AD','AR','AU','AT','BH','BY','BE','BO','BA','BR','BG','CA','CL','CO','CR','HR','CY','CZ','DK','DO','EC','EG','SV','EE','FI','FR','DE','GR','GT','HN','HK','HU','IS','IN','ID','IE','IL','IT','JP','JO','KZ','XK','KW','LV','LB','LI','LT','LU','MY','MT','MX','MD','MC','ME','MA','NL','NZ','NI','MK','NO','OM','PS','PA','PY','PE','PH','PL','PT','QA','RO','RU','SA','RS','SG','SK','SI','ZA','ES','SE','CH','TW','TH','TN','TR','UA','AE','GB','US','UY','VN'] in the search.

shillshocked:
[...]

PR #526, opened 2020-07-02
Add support to search multiple markets at once.

NLeRoy917:
Resolve issue #525 

## Overview
 - Support to search multiple markets at once by passing in a list.
 - Support to search all available Spotify markets.
 - Added two unit tests to test the new feature.

## Detailed
I reformatted the API interface `search` method. I did notice that all but one method directly returned the result that comes from the `self._get` call, and I wanted to keep this to retain any abstraction, but I couldn't think of a way around it. The new `search` method on the Spotify client keeps all original functionality, but I added in some logic to handle if a user passes in a list of markets to search through. In addition, as referenced in #525, I added in the ability to search **ALL** markets as well as another available parameter `n` that will stop the searching once a given number of results are found.

I really tried to keep the object structure returned from Spotify... So, the final object returned from `search` when multiple markets are passed in will be almost identical to the object returned if one market or no markets are passed in.

To the best of my knowledge, it is the same, except `href` now results in a list of `href`s used for searching.

## Performance Issues
As mentioned by @stephanebruckert, this code performs poorly due to looping through http requests, but I can not think of a way around it.

### DIFF
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -1583,7 +1583,7 @@ class Spotify(object):
             results[type + 's']['items'] += result[type + 's']['items']
             results[type + 's']['total'] += result[type + 's']['total']
             if total and len(results[type + 's']['items']) >= total:
-                # sice 'items' to only include number of results requested
+                # splice 'items' to only include number of results requested
                 results[type + 's']['items'] = results[type + 's']['items'][:total]
                 return results
         return results
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -523,7 +523,7 @@ class Spotify(object):
                            from_token. Can supply list of markets. Pass "ALL" to search all country codes.
                 - total - the total number of results to return if multiple markets are supplied in the search.
         """
-        # if string passed and equals "ALL"
+
         if (isinstance(market, str) and market.upper() == "ALL"):
             warnings.warn(
                 "Searching all markets is poorly performing.",
@@ -531,15 +531,13 @@ class Spotify(object):
             )
             return self._search_multiple_markets(q, limit, offset, type, self.country_codes, total)
 
-        # if list is passed
-        elif isinstance(market, list):
+        elif isinstance(market, list) or isinstance(market, tuple):
             warnings.warn(
                 "Searching multiple markets is poorly performing.",
                 UserWarning,
             )
             return self._search_multiple_markets(q, limit, offset, type, market, total)
 
-        # handle all other cases
         else:
             return self._get(
                 "search", q=q, limit=limit, offset=offset, type=type, market=market
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -508,19 +508,20 @@ class Spotify(object):
         tlist = [self._get_id("episode", e) for e in episodes]
         return self._get("episodes/?ids=" + ",".join(tlist), market=market)
 
-    def search(self, q, limit=10, offset=0, type="track", market=None, n=None):
+    def search(self, q, limit=10, offset=0, type="track", market=None, total=None):
         """ searches for an item
 
             Parameters:
                 - q - the search query (see how to write a query in the
                       official documentation https://developer.spotify.com/documentation/web-api/reference/search/search/)  # noqa
-                - limit  - the number of items to return (min = 1, default = 10, max = 50)
+                - limit  - the number of items to return (min = 1, default = 10, max = 50). If a search is to be done on multiple
+                            markets, then this limit is applied to each market. (e.g. search US, CA, MX each with a limit of 10).
                 - offset - the index of the first item to return
                 - type - the type of item to return. One of 'artist', 'album',
                          'track', 'playlist', 'show', or 'episode'
                 - market - An ISO 3166-1 alpha-2 country code or the string
-                           from_token. Can supply list of markets. Pass "ALL" for to search all country codes.
-                - n - return as soon as n options are found
+                           from_token. Can supply list of markets. Pass "ALL" to search all country codes.
+                - total - the total number of results to return if multiple markets are supplied in the search.
         """
         # if string passed and equals "ALL"
         if (isinstance(market, str) and market.upper() == "ALL"):
@@ -528,29 +529,7 @@ class Spotify(object):
                 "Searching all markets is poorly performing.",
                 UserWarning,
             )
-            results = {
-                type + 's': {
-                    'href': [],
-                    'items': [],
-                    'limit': limit,
-                    'next': None,
-                    'offset': 0,
-                    'previous': None,
-                    'total': 0
-                }
-            }
-            for country in self.country_codes:
-                result = self._get(
-                    "search", q=q, limit=limit, offset=offset, type=type, market=country
-                )
-                results[type + 's']['href'].append(result[type + 's']['href'])
-                results[type + 's']['items'] += result[type + 's']['items']
-                results[type + 's']['total'] += result[type + 's']['total']
-                if n and len(results[type + 's']['items']) >= n:
-                    return results
-
-            # raise warning
-            return results
+            return self._search_multiple_markets(q, limit, offset, type, self.country_codes, total)
 
         # if list is passed
         elif isinstance(market, list):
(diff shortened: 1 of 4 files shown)


spotipy/client.py
```diff
@@ -458,11 +519,74 @@ def search(self, q, limit=10, offset=0, type="track", market=None):
...
+                    "search", q=q, limit=limit, offset=offset, type=type, market=country
+                )
+                results[type + 's']['href'].append(result[type + 's']['href'])
+                results[type + 's']['items'] += result[type + 's']['items']
+                results[type + 's']['total'] += result[type + 's']['total']
+                if n and len(results[type + 's']['items']) >= n:
+                    return results
+
+            # raise warning
+            return results
```
stephanebruckert, line 583: This part is almost identical to the part above apart from the list of markets, any chance we can move it to a function?

spotipy/client.py
```diff
@@ -447,7 +448,7 @@ def episodes(self, episodes, market=None):
         tlist = [self._get_id("episode", e) for e in episodes]
         return self._get("episodes/?ids=" + ",".join(tlist), market=market)
 
-    def search(self, q, limit=10, offset=0, type="track", market=None):
+    def search(self, q, limit=10, offset=0, type="track", market=None, n=None):
```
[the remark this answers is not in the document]
NLeRoy917: Yeah, I like the idea of including some parameter to throttle back the market search... `limit` is used per-market, so maybe `total` parameter? Something that indicates the total number of tracks to return - I can reflect this in the docstring.

spotipy/client.py
NLeRoy917: Good idea. I can make an internal `self._search_multiple_markets` function to handle both cases

spotipy/client.py
```diff
@@ -34,6 +34,67 @@ class Spotify(object):
     """
     max_retries = 3
     default_retry_codes = (429, 500, 502, 503, 504)
+    country_codes = [
```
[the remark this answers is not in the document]
NLeRoy917: I searched for awhile for a comprehensive list of available markets supported by Spotify - I was able to find [this](https://github.com/spotify/web-api/issues/442#issuecomment-315400463) issue over on the spotify/web-api repo. It's a bummer Spotify wont provide an endpoint or at least a list of markets that can be searched, so I figured this would be next best - and it could be updated as needed.

spotipy/client.py
```diff
@@ -458,11 +519,74 @@ def search(self, q, limit=10, offset=0, type="track", market=None):
...
+                result = self._get(
+                    "search", q=q, limit=limit, offset=offset, type=type, market=country
+                )
+                results[type + 's']['href'].append(result[type + 's']['href'])
+                results[type + 's']['items'] += result[type + 's']['items']
+                results[type + 's']['total'] += result[type + 's']['total']
+                if n and len(results[type + 's']['items']) >= n:
+                    return results
+
+            # raise warning
```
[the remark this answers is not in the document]
NLeRoy917: Yes. Oops.

tests/integration/test_non_user_endpoints.py
```diff
@@ -175,6 +175,18 @@ def test_artist_search_with_market(self):
...
+        self.assertTrue('artists' in results)
+        self.assertTrue(len(results['artists']['items']) > 0)
+        self.assertTrue(results['artists']['items'][0]['name'] == 'Weezer')
+
+    def test_artist_search_with_all_markets(self):
+        results = self.spotify.search(q='weezer', type='artist', market='ALL')
+        self.assertTrue('artists' in results)
+        self.assertTrue(len(results['artists']['items']) > 0)
+        self.assertTrue(results['artists']['items'][0]['name'] == 'Weezer')
+
```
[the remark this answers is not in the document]
NLeRoy917: I added a test for this: `self.assertTrue(len(results_limited['artists']['items']) <= TOTAL)`. I have it checking for less than or equal to, since the result may not always find the total number requested.

NLeRoy917:
Will adjust formatting to try and resolve those checks in just a bit - local flake8 didn’t find anything

spotipy/client.py
```diff
@@ -1605,3 +1562,30 @@ def _get_id(self, type, id):
...
+            }
+        for country in markets:
+            result = self._get(
+                "search", q=q, limit=limit, offset=offset, type=type, market=country
+            )
+            results[type + 's']['href'].append(result[type + 's']['href'])
+            results[type + 's']['items'] += result[type + 's']['items']
+            results[type + 's']['total'] += result[type + 's']['total']
+            if total and len(results[type + 's']['items']) >= total:
+                # sice 'items' to only include number of results requested
```
stephanebruckert, line 1586: typo: slice*?

stephanebruckert:
Do you have any improvement idea on this maybe @ritiek? Otherwise really happy to merge this

stephanebruckert:
Do you have any improvement idea on this maybe @ritiek? Otherwise really happy to merge

spotipy/client.py
NLeRoy917: Yes. Will fix that one!

spotipy/client.py
```diff
@@ -447,22 +508,42 @@ def episodes(self, episodes, market=None):
...
+        # if string passed and equals "ALL"
+        if (isinstance(market, str) and market.upper() == "ALL"):
+            warnings.warn(
+                "Searching all markets is poorly performing.",
+                UserWarning,
+            )
+            return self._search_multiple_markets(q, limit, offset, type, self.country_codes, total)
+
+        # if list is passed
+        elif isinstance(market, list):
```
ritiek, line 535: Let's also accept `tuple` along with `list` here.

spotipy/client.py
```diff
@@ -447,22 +508,42 @@ def episodes(self, episodes, market=None):
...
                 - market - An ISO 3166-1 alpha-2 country code or the string
-                           from_token.
-        """
-        return self._get(
-            "search", q=q, limit=limit, offset=offset, type=type, market=market
-        )
+                           from_token. Can supply list of markets. Pass "ALL" to search all country codes.
+                - total - the total number of results to return if multiple markets are supplied in the search.
+        """
+        # if string passed and equals "ALL"
```
ritiek, line 526: This comment doesn't really tell anything different from the line of code just below it. Let's change it to something like "Search in all available markets" or maybe we could just remove it all together.

spotipy/client.py
```diff
@@ -447,22 +508,42 @@ def episodes(self, episodes, market=None):
...
+        if (isinstance(market, str) and market.upper() == "ALL"):
+            warnings.warn(
+                "Searching all markets is poorly performing.",
+                UserWarning,
+            )
+            return self._search_multiple_markets(q, limit, offset, type, self.country_codes, total)
+
+        # if list is passed
```
ritiek, line 534: Same thing about the comment here.

spotipy/client.py
```diff
@@ -447,22 +508,42 @@ def episodes(self, episodes, market=None):
...
+        elif isinstance(market, list):
+            warnings.warn(
+                "Searching multiple markets is poorly performing.",
+                UserWarning,
+            )
+            return self._search_multiple_markets(q, limit, offset, type, market, total)
+
+        # handle all other cases
```
ritiek, line 542: And here too.

spotipy/client.py
```diff
@@ -1481,3 +1562,28 @@ def _get_id(self, type, id):
...
+            result = self._get(
+                "search", q=q, limit=limit, offset=offset, type=type, market=country
+            )
+            results[type + 's']['href'].append(result[type + 's']['href'])
+            results[type + 's']['items'] += result[type + 's']['items']
+            results[type + 's']['total'] += result[type + 's']['total']
+            if total and len(results[type + 's']['items']) >= total:
+                # splice 'items' to only include number of results requested
+                results[type + 's']['items'] = results[type + 's']['items'][:total]
+                return results
```
ritiek, line 1586: It seems weird to me here that we're messing with the API response returned by Spotify. This could be confusing at times.

I think a better choice would be to have a key for each market in the result. This way we do not mess with the internal response returned by Spotify. The results would then look something like below on using `sp.search`:

```python
>>> result = sp.search("lany good guys", market=["US", "GB"])
>>> us_tracks = result["US"]["tracks"]
>>> us_tracks.keys()
dict_keys(['href', 'items', 'limit', 'next', 'offset', 'previous', 'total'])

>>> us_tracks["href"]
'https://api.spotify.com/v1/search?query=lany+good+guys&type=track&market=US&offset=0&limit=10'

>>> gb_tracks = result["GB"]["tracks"]
>>> gb_tracks["href"]
'https://api.spotify.com/v1/search?query=lany+good+guys&type=track&market=GB&offset=0&limit=10'
```

@stephanebruckert what do you think?

NLeRoy917: I completely agree - It feels hacky. I was actually thinking of implementing the market key idea you mention as I was working on it, but I wanted to try and preserve the format of the response from Spotify.

stephanebruckert: Sorry I merged this too quickly. I agree, it does feel weird. Now that you point it out it's true that mixing up results does not make sense, mainly due to the fact that results are supposed to be ordered by "relevancy".

The reason I thought it was OK in the first place is that in many cases, we are only interested in the first result. In this specific case, getting a dictionary of all countries can make it difficult to find in which country the release was found.

But it's probably nicer to let developers handle country results the way they want, since we don't really know what their use case is. So I would be all for this change

ritiek: @stephanebruckert It's all good. Thank you for maintaining spotipy all this while! I agree that devs should be able to do whatever they want with the country results later on.

@NLeRoy917 Are you willing to make a new PR for this?

NLeRoy917: Yes, definitely - I can work on that

PR #527, opened 2020-07-03
Fix return object formatting for multi-market searches

NLeRoy917:
I implemented the changes we discussed in #526. Now, the return object will be a dictionary with keys corresponding to the markets specified in the search with the response from Spotify stored in those keys. I also updated all of my unit tests to reflect these changes

### DIFF
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -1591,7 +1591,7 @@ class Spotify(object):
 
             count += len(result[first_type]['items'])
             if total and count >= total:
-                return results
+                break
             if total and limit > total - count:
                 # when approaching `total` results, adjust `limit` to not request more
                 # items than needed
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -508,7 +508,7 @@ class Spotify(object):
         tlist = [self._get_id("episode", e) for e in episodes]
         return self._get("episodes/?ids=" + ",".join(tlist), market=market)
 
-    def search(self, q, limit=10, offset=0, type="track", market=None, total=None):
+    def search(self, q, limit=10, offset=0, type="track", market=None):
         """ searches for an item
 
             Parameters:
@@ -520,7 +520,24 @@ class Spotify(object):
                 - type - the type of item to return. One of 'artist', 'album',
                          'track', 'playlist', 'show', or 'episode'
                 - market - An ISO 3166-1 alpha-2 country code or the string
-                           from_token. Can supply list of markets. Pass "ALL" to search all country codes.
+                           from_token.
+        """
+        return self._get(
+            "search", q=q, limit=limit, offset=offset, type=type, market=market
+        )
+
+    def search_markets(self, q, limit=10, offset=0, type="track", market=None, total=None):
+        """ searches multple markets for an item
+
+            Parameters:
+                - q - the search query (see how to write a query in the
+                      official documentation https://developer.spotify.com/documentation/web-api/reference/search/search/)  # noqa
+                - limit  - the number of items to return (min = 1, default = 10, max = 50). If a search is to be done on multiple
+                            markets, then this limit is applied to each market. (e.g. search US, CA, MX each with a limit of 10).
+                - offset - the index of the first item to return
+                - type - the type's of item's to return. One or more of 'artist', 'album',
+                         'track', 'playlist', 'show', or 'episode'. If multiple types are desired, pass in a comma separated list.
+                - market - A list of ISO 3166-1 alpha-2 country codes. Pass "ALL" to search all country markets.
                 - total - the total number of results to return if multiple markets are supplied in the search.
         """
 
@@ -538,6 +555,7 @@ class Spotify(object):
             )
             return self._search_multiple_markets(q, limit, offset, type, market, total)
 
+        # is this needed anymore?
         else:
             return self._get(
                 "search", q=q, limit=limit, offset=offset, type=type, market=market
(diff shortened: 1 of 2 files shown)


spotipy/client.py
```diff
@@ -1562,26 +1562,16 @@ def _get_uri(self, type, id):
...
         for country in markets:
             result = self._get(
                 "search", q=q, limit=limit, offset=offset, type=type, market=country
             )
-            results[type + 's']['href'].append(result[type + 's']['href'])
-            results[type + 's']['items'] += result[type + 's']['items']
-            results[type + 's']['total'] += result[type + 's']['total']
-            if total and len(results[type + 's']['items']) >= total:
+            results[country] = result
+            if total and len(results[country][type + 's']['items']) >= total:
```
ritiek, line 1571: Let's store `type + 's'` into a variable first and then later on use this variable here and below as required. That should improve readability and should be a bit faster as it wouldn't have to concatenate the same strings again every time.

spotipy/client.py
```diff
@@ -1562,26 +1562,16 @@ def _get_uri(self, type, id):
...
                 # splice 'items' to only include number of results requested
```
ritiek, line 1572: splice -> slice

spotipy/client.py
```diff
@@ -1562,26 +1562,16 @@ def _get_uri(self, type, id):
...
+        results = {}
         for country in markets:
             result = self._get(
                 "search", q=q, limit=limit, offset=offset, type=type, market=country
             )
-            results[type + 's']['href'].append(result[type + 's']['href'])
-            results[type + 's']['items'] += result[type + 's']['items']
-            results[type + 's']['total'] += result[type + 's']['total']
-            if total and len(results[type + 's']['items']) >= total:
+            results[country] = result
```
stephanebruckert, line 1570: Should we only add countries with results? This way listing all keys directly shows countries with results

NLeRoy917:
Yes, I agree inconsistent return result structures could be confusing or break code... this was the motivation behind my original pull request.

I think either way you might have pitfalls - but I am open to revising again. 

I’m addition I can add a check that if a list (or tuple) is passed, it can be treated as a single market entry

ritiek:
@stephanebruckert Yep, that's looks like an inconsistency too. I think we could then create a new method for searching in multiple markets like you mentioned. We keep the `search` method as it is and create a newer one. `search_markets` sounds a good one.

spotipy/client.py
```diff
@@ -1562,26 +1562,15 @@ def _get_uri(self, type, id):
...
-                    'items': [],
-                    'limit': limit,
-                    'next': None,
-                    'offset': 0,
-                    'previous': None,
-                    'total': 0
-                }
-            }
+        results = {}
+        types = type + 's'
```
stephanebruckert, line 1566: Just realised that `type` can be a comma separated list of types such as `album,track`

https://developer.spotify.com/documentation/web-api/reference/search/search/

So the type would need to be split and looped over. I checked, if we do `search(limit=4, `track,artist`), it will return 4 tracks and 4 artists.

---------------

Another possibility is to "fix" `total` at the beginning of the method so that if it exists, it is not less than limit. If `limit` > `total`, e.g. `search_markets(limit=4, total=2, markets=['US', 'GB', 'AU])`, then show a warning that says something like:
> Warning: `limit` was automatically adjusted to equal `total` as `total` must equal or be higher than `limit`.

NLeRoy917: I was unaware of type having the ability to be multiple values. I can edit this too in addition to making a new `search_market` method

stephanebruckert: Thanks a lot @NLeRoy917 🙏

NLeRoy917: No problem! Just to make sure I digested it all properly:
* Revert `search` to its original method
* Create a new `search_markets` method incorporating the changes
* Check for multiple types and loop through to incorporate all the results
* Auto-adjust limit to total value if total is less than limit

stephanebruckert: All good apart from number 3 that I think you shouldn't need anymore if you already do number 4

NLeRoy917: Right - I believe this means that total will be the total returned for each country instead of total for ALL countries.

stephanebruckert:
^ just added a commit for old code I wasn't able to comment

stephanebruckert:
Since I was there I added another quick commit to rename `market` -> `markets` and fixed the test

Unfortunately I realised that since multiple types can be requested, "total" cannot refer to all types at the same time... so I made it use the first type only for now. I'm sure this can be improved somehow but for now it's probably ok? I added that detail to the docstring

spotipy/client.py
```diff
@@ -1588,10 +1579,21 @@ def _search_multiple_markets(self, q, limit, offset, type, markets, total):
...
+
         for country in markets:
             result = self._get(
                 "search", q=q, limit=limit, offset=offset, type=type, market=country
             )
             results[country] = result
 
+            count += len(result[first_type]['items'])
+            if total and count >= total:
+                return results
```
NLeRoy917: Would we need to slice `result[first_type]['items']`? Like I was doing before? In the case that `count += len(result[first_type]['items'])` sends it over the total.

stephanebruckert: No, `limit` should never be higher than `total` thanks to auto-adjusting `limit` at the top and bottom of this method

stephanebruckert:
> we can think about how to hard limit the return results across all types

We can create a new issue for this.

Are you @NLeRoy917 and @ritiek happy if we merge the current PR?

spotipy/client.py
NLeRoy917: Right - that is correct

NLeRoy917:
All looks good to me 👍

ritiek:
Looks good to me too. Thanks @NLeRoy917!

Issue #51, opened 2015-05-22
Abstraction of caching code


mikejewell:
I'm using spotipy on Heroku as a Slack bot - which works awesomely, but I've had to store the cached token on S3. I've extended SpotifyOAuth and overridden get_cached_token and _save_token_info to do this, but it'd be useful if it was possible to do this without extensio?. This would also allow for storing the token in a DB rather than in a file.

Maybe the simplest approach could be load_cached_token and save_cached_token methods, which can be overridden to cater for different storage approaches?

arska:
I had the same problem, see #61 for my workaround

markuman:
redis as caching engine would be nice

theTechGuyy:
Has there been any update on this so it is possible to use a different matter of saving sessions but file system?
Cannot seem to find any, but I might be blind / not understand well enough

PR #528, opened 2020-07-06
Allow for total headless mode by instructing the user to open the URL in a browser.

foobuzz:
This adds a new option `open_browser` in `get_auth_response` which, if False, will simply print the authorization URL and ask the user to visit it and paste the URL is has been redirected to  (existing interactive way to get the authorization code). This allows the setup of Spotipy on a remote machine (e.g. Raspberry Pi or VPS) in a fully headless way.

`open_browser` defaults to True for backward compatibility.

### DIFF
--- a/spotipy/oauth2.py
+++ b/spotipy/oauth2.py
@@ -415,7 +415,11 @@ class SpotifyOAuth(SpotifyAuthBase):
         redirect_info = urlparse(self.redirect_uri)
         redirect_host, redirect_port = get_host_port(redirect_info.netloc)
 
-        if open_browser and redirect_host in ("127.0.0.1", "localhost") and redirect_info.scheme == "http":
+        if (
+            open_browser
+            and redirect_host in ("127.0.0.1", "localhost")
+            and redirect_info.scheme == "http"
+        ):
             # Only start a local http server if a port is specified
             if redirect_port:
                 return self._get_auth_response_local_server(redirect_port)
--- a/spotipy/oauth2.py
+++ b/spotipy/oauth2.py
@@ -374,9 +374,17 @@ class SpotifyOAuth(SpotifyAuthBase):
         except webbrowser.Error:
             logger.error("Please navigate here: %s", auth_url)
 
-    def _get_auth_response_interactive(self):
-        self._open_auth_url()
-        response = SpotifyOAuth._get_user_input("Enter the URL you were redirected to: ")
+    def _get_auth_response_interactive(self, open_browser=True):
+        if open_browser:
+            self._open_auth_url()
+            prompt = "Enter the URL you were redirected to: "
+        else:
+            url = self.get_authorize_url()
+            prompt = (
+                "Go to the following URL: {}\n"
+                "Enter the URL you were redirected to: ".format(url)
+            )
+        response = SpotifyOAuth._get_user_input(prompt)
         state, code = SpotifyOAuth.parse_auth_response_url(response)
         if self.state is not None and self.state != state:
             raise SpotifyStateError(self.state, state)
@@ -397,7 +405,7 @@ class SpotifyOAuth(SpotifyAuthBase):
         else:
             raise SpotifyOauthError("Server listening on localhost has not been accessed")
 
-    def get_auth_response(self):
+    def get_auth_response(self, open_browser=True):
         logger.info('User authentication requires interaction with your '
                     'web browser. Once you enter your credentials and '
                     'give authorization, you will be redirected to '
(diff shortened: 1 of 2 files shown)


spotipy/oauth2.py
```diff
@@ -374,9 +374,17 @@ def _open_auth_url(self):
         except webbrowser.Error:
             logger.error("Please navigate here: %s", auth_url)
 
-    def _get_auth_response_interactive(self):
-        self._open_auth_url()
-        response = SpotifyOAuth._get_user_input("Enter the URL you were redirected to: ")
+    def _get_auth_response_interactive(self, open_browser=True):
+        if open_browser:
```
stephanebruckert, line 378: Why this condition? Aren't both sides doing the same thing?

spotipy/oauth2.py
```diff
@@ -374,9 +374,17 @@ def _open_auth_url(self):
...
+            self._open_auth_url()
+            prompt = "Enter the URL you were redirected to: "
+        else:
+            url = self.get_authorize_url()
+            prompt = (
+                "Go to the following URL: {}\n"
+                "Enter the URL you were redirected to: ".format(url)
+            )
```
stephanebruckert, line 386: Why this condition? Aren't both sides doing the same thing?

foobuzz: In the case where we want to open the browser, the method `_open_auth_url` will launch the browser with the authorization URL. Otherwise, this authorization URL is printed out to stdout and the user needs to figure out how to open it by himself. In both cases we ask the user for the final URL she has been redirected to.

NLeRoy917:
I'm jumping on here because headless authentication with Spotify's API has been something I've been trying to get down for awhile... Is this PR really introducing **total** headless authentication? In both cases, user input is required. A truly headless authentication flow would be completely automated, and this is very difficult to do since Spotify's OAuth flow requires a log in with a browser. Thus, some sort of browser driver (like Selenium) is required to submit credentials and literally "click" the submit button on the page.

[As far as I know](https://developer.spotify.com/documentation/general/guides/authorization-guide/#authorization-code-flow), all but the Client Credentials Authorization Flow require some sort of user interaction - which must be automated via browser driver to truly become headless

stephanebruckert:
@NLeRoy917 you are right, "total headless" should be "browserless"

stephanebruckert:
@NLeRoy917 you are right, this is more of a "browserless" mode, or at least opening a browser on the same machine is not being forced

NLeRoy917:
I wonder if a headless mode could ever be implemented without compromising the scale of spotipy. It’s advertised as “A light weight Python library for the Spotify Web API”, but almost all browser drivers I see that have javascript capabilities require some external dependencies like chromium or webkit... they are usually heavy like Selenium. This isn’t really in the spirit of spotipy.

I’ve found a few options: 
- splinter
- Ghost.py
- phantompy
- mechanicalsoup

Some are deprecated, though, and others are hard to work with.

foobuzz:
Well it all comes down to a matter of terminology anyway, but I still think that "headless" is a proper description whereas "browserless" is not. [Headless](https://en.wikipedia.org/wiki/Headless_software) merely means that it doesn't use a graphical interface, which is the case here since all the program uses is stdout and stdin. "Browserless" would misleading since the user still has to open the URL in a browser somehow. I've updated the changelog anyway to something more factual to resolve any ambiguity.

foobuzz:
@NLeRoy917 Even using a browser driver would still need to have either the username/password or an authentication cookie of a given user to load the authorization screen for this user. If one is willing to transfer such secrets to a remote machine beforehand, why not pass the authorization step on a local machine beforehand, and then transfer the OAuth token to the remote machine.

NLeRoy917:
You make a good point. But, I think it's more flexible (and scales better) to just type my username and password as environment variables and have Selenium do the work for me, rather than run a script locally and copy and paste a key into my PaaS dashboard every time I lose my authorization and refresh token

stephanebruckert:
The point of using OAuth is that there is no middleman between the user and Spotify. Since the user only gives the password to spotify.com, even the library (in this case spotipy) cannot see the password.

It's definitely possible to hack something that allows direct password login, however it's highly disrecommended:

 - the password is eventually typed or stored outside of the user's browser -> it could potentially be found in the terminal history or in some files, whereas a browser doesn't leave such trace
 - the library and its whole tree of dependencies must be trusted, which can be a lot to ask when there is already a way to do it securely

As a user I think being redirected to spotify.com in a browser at least once is fine, particularly if the user is already signed in:

 - if using a spotipy webapp, the user can just click a link and sign in
 - if using a spotipy script, it's probably fine to get redirected to a browser only once in order to sign in

However I understand the frustration as a developer who is trying to run a script remotely and continuously. I've overcome this issue by caching the spotify token into DynamoDB ([code here](https://github.com/resident-archive/resident-archive/blob/a869b73f1f64538343be1604d43693b6165cc58a/functions/to-spotify/main.py#L129..L157)). Basically:
 - run the script locally at least once,
 - if DB has no token redirect to browser and persist the token in the DB,
 - deploy and run the script wherever,
 - there should be a token in the DB so there won't be any prompt, then refresh the old token and persist

Hopefully we can implement this into spotipy one day as part of https://github.com/plamere/spotipy/issues/51. Someone also suggested Redis!

stephanebruckert:
Thanks a lot for this @foobuzz, merging it now will close 3 issues at once!

Issue #507, opened 2020-06-01
spotipy.oauth2.SpotifyOauthError: You must either set a cache_path or a username


jcclark141152:
I am trying to set up a web app and am first following your example app.py. I am running in to the following issues:

1.) SpotifyOauthError message for not passing a redirect_uri.
This was corrected by passing a url that has also been added to the Spotify Developer Dashboard.
auth_manager = spotipy.oauth2.SpotifyOAuth()
BECOMES
auth_manager = spotipy.oauth2.SpotifyOAuth(redirect_uri='http://localhost:5000/')

At this point I am able to open the app in my localhost and click the generated "Sign In" hyperlink. It loads the Spotify Authorize page, but when I click the "Agree" button I receive the following error:

spotipy.oauth2.SpotifyOauthError
spotipy.oauth2.SpotifyOauthError: You must either set a cache_path or a username.

I don't have a given user's username for the purpose of this app, since the user is supposed to be authenticating with Spotify and granting me scopes. Any examples on stackoverflow or your app.py are not passing a username.

jcclark141152:
Issue resolved by passing a cache file destination path:
`sp_oauth = spotipy.oauth2.SpotifyOAuth(client_id=CLIENT_ID, client_secret=CLIENT_SECRET ,redirect_uri=REDIRECT_URI,scope=SCOPE, show_dialog=True, cache_path=CACHE)`

Working through how to structure the caches files/folder for multiple users to be stored, as well as how to remove/delete a cache for a given user to allow for login/log off functionality to switch users logging in to the app.

stephanebruckert:
> Working through how to structure the caches files/folder for multiple users to be stored, as well as how to remove/delete a cache for a given user to allow for login/log off functionality to switch users logging in to the app.

See if this example can help https://github.com/plamere/spotipy/blob/master/examples/app.py, it allows multiple users to sign in at the same time  Currently it doesn't clear the tokens in the back-end, but that's something that can be improved

MikeiLL:
I don't know that this issue is resolved. When, in [the App example](https://github.com/plamere/spotipy/blob/master/examples/app.py), we instantiate `spotipy.oauth2.SpotifyOAuth` without specifying either a `username` or `cache_path`, the example returns an error, which you can see on line 274, `elif not self.cache_path and not self.username` it raises `SpotifyOauthError`, "You must either set a cache_path or a username.")

[...]

Issue #363, opened 2019-03-27
warning:couldn't write token cache to .cache-xxxxxxxxxx


sifus:

sifus:
Hello, i can't go furether because my token can't go to the cache.

Is someone have a solution ...

TineshaErskine:
Did you ever figure this out?
I'm running into the same issue. I'm running code to look at user's top artist.

I get:
'**warning:couldn't write token cache to .cache-C**'

and a link to a page that states:
**{
  "error": {
    "status": 401,
    "message": "No token provided"
  }
}**

I have read that a lot of people were having trouble with the token part but I double checked and it seems right:

**'token = util.prompt_for_user_token(username, scope, client_id, client_secret, redirect_uri)
sp = spotipy.Spotify(auth=token)'**

![image](https://user-images.githubusercontent.com/45402779/55757314-a0d14280-5a21-11e9-8a0b-ce8f7c3d4a34.png)

HurleybirdJr:
I may have a solution.

### Problem:
On Windows 10, Windows Defender tends to block new programs from writing to personal C:/ areas (Documents, Pictures, etc...) with "**_Ransomware protection_**" and with proper error catching, would normally say the file/folder doesn't exist.

### Solution:
What I did was:

- Go into Windows Defender, under "**_Virus & threat protection_**" 
- Then "**_Ransomware protection_**" into "**_Allow an app through controlled folder access_**".

Assuming you're the administrator:

- Go to "**_Add an allowed app_**" then "**_Recently blocked apps_**" 
- Then a list of programs and their respective directories should appear. For me "**_pythonw.exe_**" was listed at the top.
- Then, you can go ahead and add it to the whitelist and if you rerun your program, it should work without that error popping up again...

Hope that helps :)

stephanebruckert:
@HurleybirdJr nice, we can add that to the FAQ

jane-dempsey:
I am having a similar problem myself. I am trying to run my .py script with a batch file on my command line. After it authorizes my Spotify and I enter the URl I was directed to, I get this output: 
Couldn't write token to cache at: .cache-{{my spotify username}}
Usage: {the path where my file is} {my spotify username}

stephanebruckert:
@jane-dempsey is the same happening if you try to run the python file directly, without the help of a batch file? Perhaps, the batch file doesn't have sufficient authorization and would need to be added to the list of allowed programs, or it should be ran as an administrator

[...]

PR #529, opened 2020-07-06
Add redirect_uri to the environment

MikeiLL:
And specify that it should be the address that the App runs over. 

Would it also be useful for the example to include a cache_path, so that it "works" more immediately?

```
CACHE = ".cache-" + "test"
SCOPE = None
# user-top-read user-library-read playlist-read-public user-follow-read
auth_manager = spotipy.oauth2.SpotifyOAuth(credentials.SPOTIPY_CLIENT_ID, credentials.SPOTIPY_CLIENT_SECRET, credentials.SPOTIPY_REDIRECT_URI, scope=SCOPE, show_dialog=True, cache_path=CACHE)
```

### DIFF
--- a/examples/app.py
+++ b/examples/app.py
@@ -5,6 +5,7 @@ Prerequisites
 
     export SPOTIPY_CLIENT_ID=client_id_here
     export SPOTIPY_CLIENT_SECRET=client_secret_here
+    export SPOTIPY_REDIRECT_URI='http://127.0.0.1:8080' // added to your [app settings](https://developer.spotify.com/dashboard/applications)
     // on Windows, use `SET` instead of `export`
 
 Run app.py


stephanebruckert:
Thanks for the PR.

Responding to your comment, does it not work without specifying CACHE? I think it currently stores a cache file for each user, which is useful because that's what remembers each token

MikeiLL:
>  does it not work without specifying CACHE? I think it currently stores a cache file for each user, which is useful because that's what remembers each token

Correct. Looks like `spotify.get_cached_token()` raises `"You must either set a cache_path or a username."` unless either a cache path or username is specified, which is what I'm seeing in the browser (at least with `export FLASK_ENV=development` set).

And as other users have commented, if we already have a username, we probably don't need to interactively Auth in the first place.

stephanebruckert:
I'm going to merge this since adding the REDIRECT_URI is useful.

However could please link to the issue and comments you are referring to? it doesn't seem related to the Flask example, am I right?

MikeiLL:
> I'm going to merge this since adding the REDIRECT_URI is useful.

Great.

> Edit: are you using Windows? #363

No. I'm too lazy to use Windows. OSX.

stephanebruckert:
About "You must either set a cache_path or a username.", I'm not sure what bug you are referring to. Please create a new issue or refer to an existing one

MikeiLL:
> About "You must either set a cache_path or a username.", I'm not sure what bug you are referring to. Please create a new issue or refer to an existing one

It's this one, I think, that applies best: https://github.com/plamere/spotipy/issues/507

stephanebruckert:
@MikeiLL you were right, I double checked and we shouldn't have to specify a username... It doesn't make sense to specify a username when we don't know yet who is going to connect.

Unfortunately setting a cache `CACHE = ".cache-" + "test"` (or equally specifying a username) won't allow multiple users to use the app safely and without conflicts. It means that this script does not do what it is supposed to do.

I'm going to work on a solution, but this might take a bit of time and might even require a version 3 for the library, as most of the caching stuff is old and would need to be removed, making some functions backward-incompatible.

So thank you for the report, I really thought I once saw it working properly for multiple users but it looks like it's not the case!

MikeiLL:
One workaround for multiple, non-concurrent, users that seems to be working is to delete the cache file on logout:

```
CACHE = ".cache-" + "test"

auth_manager = spotipy.oauth2.SpotifyOAuth(credentials.SPOTIPY_CLIENT_ID, 
									credentials.SPOTIPY_CLIENT_SECRET, 
									credentials.SPOTIPY_REDIRECT_URI, 
									scope='user-read-currently-playing', 
									show_dialog=True, 
									cache_path=CACHE)

# code truncated

@app.route('/sign_out')
def sign_out():
	session.clear()
	try:
		os.remove(CACHE)
	except OSError as e:  ## if failed, report it back to the user ##
		print ("Error: %s - %s." % (e.filename, e.strerror))
	return redirect('/')
```

Are you looking at support for multiple concurrent users?

stephanebruckert:
@MikeiLL this will work if user B signs in after user A signs out. However might cause issues if both users use the app at the same time / AKA user the same cache file

I have a work in progress on this branch https://github.com/plamere/spotipy/pull/539, it seems to work apart from refreshing tokens which I will add later. But that's something you could already try in case you need this asap

PR #539, opened 2020-07-12
Fix Flask example app

stephanebruckert:
fixed:

 - no username or path needed
 - can be used by multiple users

### DIFF
--- a/examples/app.py
+++ b/examples/app.py
@@ -22,24 +22,34 @@ import uuid
 app = Flask(__name__)
 app.config['SECRET_KEY'] = os.urandom(64)
 app.config['SESSION_TYPE'] = 'filesystem'
-
+app.config['SESSION_FILE_DIR'] = './.flask_session/'
 Session(app)
+
+caches_folder = './.spotify_caches/'
+if not os.path.exists(caches_folder):
+    os.makedirs(caches_folder)
 
 
 @app.route('/')
 def index():
-    cache_path = '.cache-'.join(str(uuid.uuid4()))
-    auth_manager = spotipy.oauth2.SpotifyOAuth(cache_path=cache_path)
+    if not session.get('uuid'):
+        # Step 1. Visitor is unknown, give random ID
+        session['uuid'] = str(uuid.uuid4())
+
+    auth_manager = spotipy.oauth2.SpotifyOAuth(cache_path=session_cache_path(session.get('uuid')))
+
     if request.args.get("code"):
-        session['token_info'] = auth_manager.get_access_token(request.args["code"], check_cache=False)
+        # Step 3. Being redirected from Spotify auth page
+        session['token_info'] = auth_manager.get_access_token(request.args.get("code"))
         return redirect('/')
 
-    token_info = session.get('token_info')
-    if not token_info:
+    if not session.get('token_info'):
+        # Step 2. Display sign in link when no token
         auth_url = auth_manager.get_authorize_url()
         return f'<h2><a href="{auth_url}">Sign in</a></h2>'
 
-    spotify = spotipy.Spotify(auth=token_info['access_token'])
+    # Step 4. Signed in, display data
+    spotify = spotipy.Spotify(auth=session.get('token_info')['access_token'])
     return f'<h2>Hi {spotify.me()["display_name"]}, ' \
            f'<small><a href="/sign_out">[sign out]<a/></small></h2>' \
            f'<a href="/playlists">my playlists</a>'
@@ -47,6 +57,7 @@ def index():
 
 @app.route('/sign_out')
 def sign_out():
+    os.remove(session_cache_path(session.get('uuid')))
     session.clear()
     return redirect('/')
 
@@ -54,7 +65,13 @@ def sign_out():
 @app.route('/playlists')
 def playlists():
     token_info = session.get('token_info')
+
     if not token_info:
         return redirect('/')
+
     spotify = spotipy.Spotify(auth=token_info['access_token'])
     return spotify.current_user_playlists()
+
+
+def session_cache_path(session_uuid):
+    return caches_folder + session_uuid
--- a/examples/app.py
+++ b/examples/app.py
@@ -3,9 +3,11 @@ Prerequisites
 
     pip3 install spotipy Flask Flask-Session
 
+    // from your [app settings](https://developer.spotify.com/dashboard/applications)
     export SPOTIPY_CLIENT_ID=client_id_here
     export SPOTIPY_CLIENT_SECRET=client_secret_here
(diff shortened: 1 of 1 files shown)


stephanebruckert:
@ritiek I would be interested by any improvement you might have on this, as many users rely on such example to get started. Does it look secure enough to you?

ritiek:
@stephanebruckert Sure. But for some reason I can't get this to work correctly I think.

I first set the env:
```console
export SPOTIPY_CLIENT_ID=abc123....
export SPOTIPY_CLIENT_SECRET=xyz789...
export SPOTIPY_REDIRECT_URI=http://localhost:8500/callback
export FLASK_APP=/path/to/spotipy/examples/app.py
```

and then launched the example with:
```console
$ python3 -m flask run --port=8080
```

Then I went to `127.0.0.1:8080` on my browser and which showed the "Sign in". On clicking that I get redirected to the Spotify OAuth login page, which in turn redirects me to `SPOTIPY_REDIRECT_URI` with the callback code appended in the URL after I authorize the OAuth login. I'm not sure what should happen after this.

If I now go to `127.0.0.1:8080` again in my web browser it still shows the "Sign in" link which does the same thing as above.

If I go to `127.0.0.1:8080/playlists`, I get redirected back to `127.0.0.1:8080` (still showing the "Sign in" link). If I go to `127.0.0.1:8080/sign_out`, I get this traceback in the console:
```bash
127.0.0.1 - - [24/Jul/2020 16:58:36] "GET / HTTP/1.1" 200 -
[2020-07-24 16:58:40,136] ERROR in app: Exception on /sign_out [GET]
  File "/home/ritiek/.local/lib/python3.8/site-packages/flask/app.py", line 2447, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/ritiek/.local/lib/python3.8/site-packages/flask/app.py", line 1952, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/ritiek/.local/lib/python3.8/site-packages/flask/app.py", line 1821, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/ritiek/.local/lib/python3.8/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/home/ritiek/.local/lib/python3.8/site-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/ritiek/.local/lib/python3.8/site-packages/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/ritiek/Downloads/spotipy/examples/app.py", line 61, in sign_out
    os.remove(session_cache_path(session.get('uuid')))
FileNotFoundError: [Errno 2] No such file or directory: './.spotify_caches/9174cd42-ad7b-47d4-baf3-1495f2512ff5'
127.0.0.1 - - [24/Jul/2020 16:58:40] "GET /sign_out HTTP/1.1" 500 -
```
and Internal Server Error on the webpage: "The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application."

MikeiLL:
> ```
> SPOTIPY_REDIRECT_URI=http://localhost:8500/callback

I believe that the `SPOTIPY_REDIRECT_URI` should be `http://127.0.0.1:8080`. And that needs to also have been added to your spotify app.

Also, if you run, `ls -a` in `/home/ritiek/Downloads/spotipy/examples`, once the app has successfully started, do you see a `.spotify_caches` directory? What's in it?

ritiek:
> I believe that the `SPOTIPY_REDIRECT_URI` should be `http://127.0.0.1:8080`. And that needs to also have been added to your spotify app.

Yup, that works! It wasn't obvious enough to me that the callback URL should be the same as where flask listens on (perhaps we should add a comment about this in the example source, others could be confused too). I can now sign in properly in the example app now and access playlists response.

> Also, if you run, `ls -a` in `/home/ritiek/Downloads/spotipy/examples`, once the app has successfully started, do you see a `.spotify_caches` directory? What's in it?

I believe you meant the current working directory I launched the example from (not necessarily `/home/ritiek/Downloads/spotipy/examples`). When running the example with the updated redirect URI, I see a file containing token information in JSON format in `.spotify_caches`. With the previous redirect URI in https://github.com/plamere/spotipy/pull/539#issuecomment-663496151 (which didn't work), the files are still created but they seem to contain binary data instead of JSON.

MikeiLL:
> > I believe that the `SPOTIPY_REDIRECT_URI` should be `http://127.0.0.1:8080`. And that needs to also have been added to your spotify app.
> 
> Yup, that works! It wasn't obvious enough to me that the callback URL should be the same as where flask listens on (perhaps we should add a comment about this in the example source, others could be confused too). 

Yes that's in [#540](https://github.com/plamere/spotipy/pull/540/commits/b8efb5a9fe365d8d8e1424f359c10845b89e5826) which we are working on.

> With the previous redirect URI in [#539 (comment)](https://github.com/plamere/spotipy/pull/539#issuecomment-663496151) (which didn't work), the files are still created but they seem to contain binary data instead of JSON.

That one isn't working for me either.

ritiek:
@stephanebruckert The example looks secure enough to me.

BTW should we just add something like `app.run(threaded=True, debug=True, host='127.0.0.1', port=8080)` at the end of example? This should make it possible to execute the example by simply calling it with python: `python app.py`.

MikeiLL:
> @stephanebruckert The example looks secure enough to me.
> 
> BTW should we just add something like `app.run(threaded=True, debug=True, host='127.0.0.1', port=8080)` at the end of example? This should make it possible to execute the example by simply calling it with python: `python app.py`.

I had added, but hadn't PR'd:

```
if __name__ == '__main__':
	app.run(port=int(os.environ.get("PORT", 8080)))
```

Maybe we should add `threaded=True` to that, but for running in debug mode, suggest at the top,

For debugging in the local environment, `export FLASK_ENV=development`.

stephanebruckert:
@MikeiLL nice catch about the port in the URL, I've just added a comment saying it was needed.

----------

> With the previous redirect URI in #539 (comment) (which didn't work), the files are still created but they seem to contain binary data instead of JSON. @ritiek 

The token info were saved as part of the session, but I've just managed to fix that! All that is saved in the session now is the uuid that's useful for spotipy to find the cached token path.

----------

> BTW should we just add something like app.run @ritiek 

> I had added, but hadn't PR'd @MikeiLL 

Sure, I'll let @MikeiLL commit that then

----------

Thanks a lot of your inputs. How does it look now? Looking forward to merge it so we can see how this works on @MikeiLL's PR https://github.com/plamere/spotipy/pull/540

stephanebruckert:
Just added `show_dialog=True` so that it always shows the spotify auth page and allows switching account

stephanebruckert:
Going to merge this, let's continue the discussion on https://github.com/plamere/spotipy/pull/540

PR #540, opened 2020-07-12
A few updates to the example app.

MikeiLL:
I added a new file, which shows an example of importing credentials from a separate file.

Hopefully this won't be messy as I see there is already a work in progress ([#539](https://github.com/plamere/spotipy/pull/539)) which I hadn't noticed.

Having multiple scopes works in the minimal example, so will have to see what's different in the app I'm developing that's consistently giving the `OSError: [Errno 48] Address already in use` error when a second scope is added.


examples/import_credentials.py
```diff
@@ -0,0 +1,28 @@
+'''
+As an alternative to exporting your credentials into the environment, 
+you may chose to create a separate file, next to your main file, 
+called credentials.py containing the following values:
```
stephanebruckert, line 4: I don't think this is a method to recommend to handle secrets, because:

 - a quick accident could happen and this file would be committed and the secrets made public,
 - if the file is missing, continuous integration might not like it

In general, prefer something like dotenv https://pypi.org/project/python-dotenv

examples/app.py
```diff
@@ -24,7 +24,12 @@
 
 Session(app)
 
-auth_manager = spotipy.oauth2.SpotifyOAuth()
+CACHE = ".cache-" + "test"
```
stephanebruckert, line 27: This will look like it works but in fact if user B signs in right after user A, user B will see A's data.

https://github.com/plamere/spotipy/pull/539 should fix multi-users usage, so let's not worry about that in the current PR

For now you can remove this line

examples/app.py
```diff
@@ -40,12 +45,19 @@ def index():
...
 
 
 @app.route('/sign_out')
 def sign_out():
     session.clear()
+    try:
+        # Remove the CACHE file (.cache-test) so that a new user can authorize.
+        os.remove(CACHE)
+    except OSError as e:
+        print ("Error: %s - %s." % (e.filename, e.strerror))
```
stephanebruckert, line 60: This can be removed as well (according to comment above)

examples/import_credentials.py
stephanebruckert: As it's not linked to the file above, feel free to either remove the file, or to add the dotenv alternative somewhere in the doc. Personally I would let developers manage the secrets how they want it

MikeiLL: That makes sense. I will remove the file.

MikeiLL: Done.

stephanebruckert:
I think I got it to work in a secure way and for multiple users https://github.com/plamere/spotipy/pull/540. If you get a chance please test it.

We would need to merge that first, and then your PR which adds `/currently_playing` and `/current_user` 👍

examples/app.py
```diff
@@ -55,3 +67,22 @@ def playlists():
         return redirect('/')
     else:
         return spotify.current_user_playlists()
+
+@app.route('/currently_playing')
+def currently_playing():
+    if not session.get('token_info'):
+        return redirect('/')
+    else:
```
stephanebruckert, line 75: since there is a `return` in the `if`, you can remove the `else`

examples/app.py
```diff
@@ -55,3 +67,22 @@ def playlists():
...
+        if not track is None:
+            return track
+        else:
+            return "No track currently playing."
+
+@app.route('/current_user')
+def current_user():
+	if not session.get('token_info'):
+		return redirect('/')
+	else:
```
stephanebruckert, line 86: same

examples/app.py
```diff
@@ -55,3 +67,22 @@ def playlists():
...
+
+@app.route('/currently_playing')
+def currently_playing():
+    if not session.get('token_info'):
+        return redirect('/')
+    else:
+        track = spotify.current_user_playing_track()
+        if not track is None:
+            return track
+        else:
```
stephanebruckert, line 79: same

examples/app.py
```diff
@@ -55,3 +67,22 @@ def playlists():
...
+            return "No track currently playing."
```
stephanebruckert, line 80: This is equivalent to:

    return track or "No track currently playing."

stephanebruckert, line 80: This is equivalent to:

    return spotify.current_user_playing_track() or "No track currently playing."

MikeiLL:
> #540



> I think I got it to work in a secure way and for multiple users #540. If you get a chance please test it.
> 
> We would need to merge that first, and then your PR which adds `/currently_playing` and `/current_user` 👍

Are you talking about [pr#539](https://github.com/plamere/spotipy/pull/539/) with `uuid.uuid4()`?

stephanebruckert:
Oops, yes https://github.com/plamere/spotipy/pull/539

MikeiLL:
> Oops, yes #539

So far it's behaving quite unexpectedly... or at least not what we want. I have tried a couple of separate `venv`s and in both instances, I never see the Spotify Auth page. I click the "login" link (which points to the Spotify page) and am automatically "logged in", so it's impossible to log in as a different user.

 I tried deleting the `.flask_session` directory. Will poke at it further when I have a chance. Thought perhaps it was because I was already logged into Spotify in the browser, but behaved the same in three different browsers.

stephanebruckert:
> I never see the Spotify Auth page. I click the "login" link (which points to the Spotify page) and am automatically "logged in", so it's impossible to log in as a different user.

I just fixed that in https://github.com/plamere/spotipy/pull/539/commits/6ed4dd839ba28d090affdd3f7833efbdb25d3528

MikeiLL:
I think I have successfully rebased the latest working multi-user example over and updated my proposed changes.

What does `Session(app)` do? Is it necessary? The code seems to work without it.

TODO: Is there a way we can define `auth_manager` and `spotify` once as opposed to in every route function? A session is required for `session_cache_path()` to work, at least as currently written.

stephanebruckert:
Just merged the other PR, you'll need to fix conflicts.

> What does Session(app) do? Is it necessary? The code seems to work without it.

`Session(app)` looks needed by flask-session https://flask-session.readthedocs.io/en/latest/#quickstart. Need to check if multiple concurrent sessions still work properly without it.

> TODO: Is there a way we can define auth_manager and spotify once as opposed to in every route function?

Yes good point, I was also thinking that was redundant. I can't think of an obvious way to do it currently. Perhaps spotipy could be updated to allow it in some way

MikeiLL:
> > TODO: Is there a way we can define auth_manager and spotify once as opposed to in every route function?
> 
> Yes good point, I was also thinking that was redundant. I can't think of an obvious way to do it currently. Perhaps spotipy could be updated to allow it in some way

That would be great. In the meantime, I wonder if the `/` route could contain a call to a function in the global scope that returns `spotify` object (maybe also the `auth_manager` obj), or `False if not session.get('uuid')`, in which case other route would be redirected back to `/`.

Another think I briefly noticed, but ignored, is that if I try to go to `logout` following the initial running of the app, it returns an exception. I forget which one at the moment. Not sure if wrapping it up in a `try/except` makes sense here.

examples/app.py
```diff
@@ -53,26 +66,58 @@ def index():
...
+		   f'<a href="/current_user">me</a>' \
 
 
 @app.route('/sign_out')
 def sign_out():
     os.remove(session_cache_path())
     session.clear()
+    try:
+        # Remove the CACHE file (.cache-test) so that a new user can authorize.
+        os.remove(CACHE)
```
stephanebruckert, line 80: this needs to be `os.remove(session_cache_path())`

MikeiLL: Done.

stephanebruckert:
> Another think I briefly noticed, but ignored, is that if I try to go to logout following the initial running of the app, it returns an exception. I forget which one at the moment. Not sure if wrapping it up in a try/except makes sense here.

Probably because it's trying to look for the `uuid` in the session and it's not there. We could look whether it's present and redirect to `/` directly if it's not.

> That would be great. In the meantime, I wonder if the / route could contain a call to a function in the global scope that returns spotify object (maybe also the auth_manager obj), or False if not session.get('uuid'), in which case other route would be redirected back to /.

Yeah definitely some improvements could be made, but I wouldn't worry too much about them if these are not so much related to spotipy. I think this fully functional example is already very helpful to anyone trying to integrate it in an API.

Merging, thanks a lot!

Issue #538, opened 2020-07-12
Implement Authorization Code Flow with Proof Key for Code Exchange (PKCE)


laymonage:
**Is your feature request related to a problem? Please describe.**
The implicit grant flow is no longer a recommended option for client-side authentication.

**Describe the solution you'd like**
[Spotify now supports PKCE](https://developer.spotify.com/community/news/2020/06/18/pkce-oauth/), so I think this library should implement it.

**Describe alternatives you've considered**
None.

**Additional context**
[Spotify PKCE auth flow guide](https://developer.spotify.com/documentation/general/guides/authorization-guide/#authorization-code-flow-with-proof-key-for-code-exchange-pkce).

stephanebruckert:
Closed by https://github.com/plamere/spotipy/pull/542

IdmFoundInHim:
@stephanebruckert
Could we get this in a release soon? Unless you're waiting for something specific, it seems like 2.15 has plenty of changes to warrant a release.

PR #542, opened 2020-07-14
Adding Initial PKCE Auth Flow Support

DJSdev:
Adds functionality for the PKCE Authorization Flow. This is now the preferred method for developers to integrate desktops and mobile applications with the Spotify API.

The benefits include:
1. Being able to distribute applications without revealing the apps Client Secret
2. Not needing users to sign up for a developer account
3. Moves away from the deprecated Implicit Grant Auth Flow 

Requested in #538 - Sorry if I jumped the gun adding this functionality without commenting on the enhancement request/issue. I needed this functionality for an app I'm working on and saw the recent enhancement request for so decided to work on it.

### DIFF
--- a/spotipy/oauth2.py
+++ b/spotipy/oauth2.py
@@ -7,6 +7,7 @@ __all__ = [
     "SpotifyOauthError",
     "SpotifyStateError",
     "SpotifyImplicitGrant",
+    "SpotifyPKCE"
 ]
 
 import base64
--- a/spotipy/oauth2.py
+++ b/spotipy/oauth2.py
@@ -578,8 +578,9 @@ class SpotifyPKCE(SpotifyAuthBase):
         self.requests_timeout = requests_timeout
         self._code_challenge_method = "S256" # Spotify requires SHA256
         self.code_verifier = self._get_code_verifier()
+        print(self.code_verifier + str(len(self.code_verifier)))
         self.code_challenge = self._get_code_challenge(self.code_verifier)
-        self.authorization_code = self.get_authorization_code()
+        self.authorization_code = None
 
     def _normalize_scope(self, scope):
         if scope:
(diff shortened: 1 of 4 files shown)


spotipy/oauth2.py
```diff
@@ -552,6 +553,369 @@ def _add_custom_values_to_token_info(self, token_info):
...
+
+    def get_access_token(self, as_dict=False, check_cache=True):
+        """ Gets the access token for the app given the code
+
+            Parameters:
+                - auth_code - the response from get_authorization_code()
+                - as_dict - a boolean indicating if returning the access token
+                            as a token_info dictionary, otherwise it will be returned
+                            as a string.
+        """
```
IdmFoundInHim, line 814: Since it is deprecated, you can remove `as_dict` from new features to discourage its continued use

IdmFoundInHim:
PKCE is the superior method for any Implicit Flow use cases, so we should follow OAuth and deprecate `SpotifyImplicitGrant`. Even though it is probably not in widespread use, I presume we cannot remove it now.

spotipy/oauth2.py
DJSdev: Removed `as_dict` as an param for `get_access_token()`

stephanebruckert:
Thanks for that! 

I agree with @IdmFoundInHim, we should probably remove `SpotifyImplicitGrant` now. Then we can merge it

IdmFoundInHim:
If we are concerned about backwards compatability, we could have `SpotifyImplicitGrant` use PKCE internally and deprecate it. (Or, we could pull the 2.13 release and group the other features into 2.14.) Otherwise, just remove the whole class and tests.

stephanebruckert:
True, so let's keep it for now as it's still present in the list of possible flows here https://developer.spotify.com/documentation/general/guides/authorization-guide/#authorization-flows

Since it's not recommended anymore, we can still deprecate it later.

Merging! Thanks @DJSdev

PR #543, opened 2020-07-17
Give a reason for player errors in SpotifyException

Esamanoaz:
I added a reason for player errors returned by Spotify. The list of reasons can be viewed here: [list](https://developer.spotify.com/documentation/web-api/reference/object-model/#player-error-reasons). This change to SpotifyException will speed up the development process by giving more specific error descriptions to developers.

An example of how this could be used:
https://pastebin.com/hLMtjMBj

### DIFF
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -240,7 +240,6 @@ class Spotify(object):
                 msg = response.json()["error"]["message"]
             except (ValueError, KeyError):
                 msg = "error"
-            
             try:
                 reason = response.json()["error"]["reason"]
             except (ValueError, KeyError):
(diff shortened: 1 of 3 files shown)


spotipy/client.py
```diff
@@ -240,6 +240,10 @@ def _internal_call(self, method, url, payload, params):
                 msg = response.json()["error"]["message"]
             except (ValueError, KeyError):
                 msg = "error"
+            try:
+                reason = response.json()["error"]["reason"]
```
stephanebruckert, line 244: we could re-use the same `response.json()` as above, then check `if "error" in json_response`. And from that `error`, populate `msg` and `reason` accordingly

Esamanoaz: That looks good!

PR #544, opened 2020-07-21
Advertise preferred language to Spotify

koukaipan:
Some artist or track provide different translation for different language
usage. This requires send http request with Accept-Language in header.
Create this option for people who wants to see proper translation.

The result showed in the following:
image

### DIFF
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -109,6 +109,7 @@ class Spotify(object):
         retries=max_retries,
         status_retries=max_retries,
         backoff_factor=0.3,
+        language=None,
     ):
         """
         Creates a Spotify API client.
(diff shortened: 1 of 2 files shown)


spotipy/client.py
```diff
@@ -224,6 +229,9 @@ def _internal_call(self, method, url, payload, params):
             if payload:
                 args["data"] = json.dumps(payload)
 
+        if self.language is not None:
+            headers["Accept-Language"] = '%s;q=1' % self.language
```
stephanebruckert, line 233: I'm trying to find what `%s;q=1` is.  Any chance you can add a very short comment to give a reference?

stephanebruckert:
Oh, also please update the changelog

spotipy/client.py
koukaipan: The q value indicates the relative priority if multiple language are specified, but since only one language is used here, I think I'd better remove it for readability. If you think this is okay, I will send a revised version and include the changelog as well.

```suggestion
            headers["Accept-Language"] = self.language
```

stephanebruckert: alright, yes it looks good without

PR #546, opened 2020-07-24
Added the parameter collaborative to user_playlist_create and "collaborative" key into data

BozeBro:
I saw that in [Create a playlist Spotify](https://developer.spotify.com/documentation/web-api/reference/playlists/create-playlist/), there was a parameter for collaborative, but the user_playlist_create() does not have collaborative as a parameter. 

I placed the collaborative key behind description because that is the order in the link.

### DIFF
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -754,9 +754,9 @@ class Spotify(object):
                 - description - the description of the playlist
         """
         data = {
-            "name": name, 
-            "public": public, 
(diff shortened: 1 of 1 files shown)


BozeBro:
Will do. I didn't know if this was considered a major enough change to put in the log at the time

stephanebruckert:
Thank you, I also merged https://github.com/plamere/spotipy/pull/548

Issue #493, opened 2020-05-17
Error getting token info - JSONDecodeError: Expecting value: line 1 column 1 (char 0)


tahsinac:
Hi, I've been using this library for a month daily on a simple script with no issues up until just now. I get this error now 

```
C:\projects\nplyrbot>"C:\ProgramData\Anaconda3\python.exe" "C:\projects\nplyrbot\nplyrbotpy.py"
Traceback (most recent call last):
  File "C:\projects\nplyrbot\nplyrbotpy.py", line 99, in <module>
    lyrics_current, current = nplyrbot()
  File "C:\projects\nplyrbot\nplyrbotpy.py", line 8, in nplyrbot
    token = util.prompt_for_user_token(username, scope, client_id=client_id, client_secret=client_secret, redirect_uri=redirect_uri)
  File "C:\ProgramData\Anaconda3\lib\site-packages\spotipy\util.py", line 88, in prompt_for_user_token
    token_info = sp_oauth.get_cached_token()
  File "C:\ProgramData\Anaconda3\lib\site-packages\spotipy\oauth2.py", line 253, in get_cached_token
    token_info = json.loads(token_info_string)
  File "C:\ProgramData\Anaconda3\lib\json\__init__.py", line 354, in loads
    return _default_decoder.decode(s)
  File "C:\ProgramData\Anaconda3\lib\json\decoder.py", line 339, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "C:\ProgramData\Anaconda3\lib\json\decoder.py", line 357, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

Not sure what's causing this all of a sudden, there's been no change in the script.

tahsinac:
Also, i have the latest version of the package installed.

stephanebruckert:
@tahsinac I see you are using Windows. Can you please check if this https://github.com/plamere/spotipy/issues/470#issuecomment-612671070 helps

tahsinac:
Hi there, just tried it and I get a browser pop up sawing '_INVALID_CLIENT: Invalid redirect URI_'. Strange as it was working just fine all this time and nothing has really changed. I went back to the Spotify web API dashboard and double checked my redirect URI. _http://localhost:8888/callback/_ is whitelisted.

tahsinac:
Follow-up:
[...]

PR #557, opened 2020-08-17
Fix JSONDecodeError on reading cache if you somehow have empty cache

Callahan633:
I added a check to see if the cache file is an empty string, as this could result in an error in `json.loads(token_info_string)` on line 287
Related open issue is #493

### DIFF
--- a/spotipy/oauth2.py
+++ b/spotipy/oauth2.py
@@ -286,7 +286,6 @@ class SpotifyOAuth(SpotifyAuthBase):
                 f.close()
                 if token_info_string != '':
                     token_info = json.loads(token_info_string)
-                token_info = ''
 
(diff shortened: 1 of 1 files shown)


stephanebruckert:
Hey @Callahan633, thanks for that. Do you know of a scenario where the token file becomes empty by any chance? we would want to fix that

It still doesn't hurt to merge your change. What do you think of adding a small warning to tell us that "something is not so normal"? for example:

    logger.warning('Ignoring current cached token file as it is empty')

Also feel free to add a short entry to the changelog https://github.com/plamere/spotipy/blob/master/CHANGELOG.md

stephanebruckert:
closing this for now as we can refer to it again via https://github.com/plamere/spotipy/issues/493

PR #559, opened 2020-08-24
Fix playlist add items

plamere:
The playlist_add_items method did not work if the given items were episodes. This was because all uris were being converted to track uris. I've changed this so that URIs are no longer automatically converted to track uris.

Also, in the unittests, the track uris and ids for 'Creep' were old causing the 'popularity' test to fail. I changed the URI to a more recent one to fix the tests.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,7 @@ extra_reqs = {
 
 setup(
     name='spotipy',
-    version='2.13.0',
+    version='2.13.1',
     description='A light weight Python library for the Spotify Web API',
     long_description=long_description,
     long_description_content_type="text/markdown",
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -18,7 +18,7 @@
 
 setup(
     name='spotipy',
-    version='2.13.0',
+    version='2.13.1',
```
stephanebruckert, line 21: we can keep this as 2.13.0 for now as we will bump it in a specific PR (2.14.0 probably soon)

stephanebruckert (approved):
Looking good

stephanebruckert:
oh, also feel free to add a quick entry to the changelog

PR #566, opened 2020-08-29
Refinement of PKCE

IdmFoundInHim:
The only effective change is that **PKCE handshake tokens will have been generated if an authorization URL is constructed**. This is important because it avoids an unexpected error from Spotify.

Other changes include:
* New public method `SpotifyPKCE.parse_auth_response_url`, simply executes the parallel method from `SpotifyOAuth`
* `SpotifyPKCE._get_auth_response_interactive` now calls a static method from `self` rather than `SpotifyOAuth`
* The `SpotifyImplicitGrant` security advisory now reflects the addition of `SpotifyPKCE`

@stephanebruckert I am [prepared to add a warning](https://github.com/IdmFoundInHim/spotipy/pull/27) to the initialization function of `SpotifyImplicitGrant`. Do we want to put it in, or just leave the advisory in the docstr?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-// Add your changes here and then delete this line
+### Added
+
+- `SpotifyPKCE.parse_auth_response_url`, mirroring that method in
+ `SpotifyOAuth`
+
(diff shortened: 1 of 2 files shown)


stephanebruckert:
Looks good to me, merging!

Yes adding the warning is a good idea https://github.com/IdmFoundInHim/spotipy/pull/27, please open that PR here 👍

Issue #549, opened 2020-07-26
I get error spotipy.oauth2.SpotifyOauthError: You must either set a cache_path or a username.


mider111:
**Describe the bug**
When I try to run my script I get the following error: spotipy.oauth2.SpotifyOauthError: You must either set a cache_path or a username.

**Your code**
    scope = 'playlist-modify-public'
    sp = spotipy.Spotify(auth_manager=SpotifyOAuth(scope=scope))
    sp.user_playlist_create(username_id, 'YouTube Converter PL', public=True, description='Test')

**Expected behavior**
I want to create a playlist. I tried calling other methods with sp but get the same error for all.

**Output**
File "D:/MyProjects/YTtoSpotify/convert.py", line 58, in create_spotify_playlist
    sp.user_playlist_create(username_id, 'YouTube Converter PL', public=True, description='Test')
  File "C:\Users\Acer\AppData\Roaming\Python\Python38\site-packages\spotipy\client.py", line 628, in user_playlist_create
    return self._post("users/%s/playlists" % (user,), payload=data)
  File "C:\Users\Acer\AppData\Roaming\Python\Python38\site-packages\spotipy\client.py", line 215, in _post
    return self._internal_call("POST", url, payload, kwargs)
  File "C:\Users\Acer\AppData\Roaming\Python\Python38\site-packages\spotipy\client.py", line 154, in _internal_call
    headers = self._auth_headers()
  File "C:\Users\Acer\AppData\Roaming\Python\Python38\site-packages\spotipy\client.py", line 145, in _auth_headers
    token = self.auth_manager.get_access_token(as_dict=False)
  File "C:\Users\Acer\AppData\Roaming\Python\Python38\site-packages\spotipy\oauth2.py", line 448, in get_access_token
    token_info = self.get_cached_token()
  File "C:\Users\Acer\AppData\Roaming\Python\Python38\site-packages\spotipy\oauth2.py", line 275, in get_cached_token
    raise SpotifyOauthError(
spotipy.oauth2.SpotifyOauthError: You must either set a cache_path or a username.

**Environment:**
 - OS: Windows 10
 - Python version 3.8
 - spotipy version 2.13
 - your IDE (if using any) PyCharm
- Have all the variables needed set up as env variables.

mider111:
Managed to go around this by passing in cache_path and redirect_uri:
`scope = 'playlist-modify-public'
    OAuth = SpotifyOAuth(scope=scope,
[...]

Issue #533, opened 2020-07-07
Spotify.OAuth Object not prompting user authorization (version 2.13.0)


a-camarillo:
I'm currently trying to run the example for the Authorization Code Flow from the docs:
```  
import spotipy
from spotipy.oauth2 import SpotifyOAuth

scope = "user-library-read"

sp = spotipy.Spotify(auth_manager=SpotifyOAuth(scope=scope))

results = sp.current_user_saved_tracks()
for idx, item in enumerate(results['items']):
    track = item['track']
    print(idx, track['artists'][0]['name'], " – ", track['name']) 
```
I can create the Spotify API client object but `SpotifyOAuth` never prompts user access. 

When I go to run `results = sp.current_user_saved_tracks()` I get the error:`spotipy.oauth2.SpotifyOauthError: You must either set a cache_path or a username.`

I'm not sure if the issue is with the Authorization Code Flow or something on my end.

a-camarillo:
I was able to prompt user authorization by calling `get_access_token` and specifying a cache_path although I'm not entirely sure how this is processing since the path is to an empty text file which doesn't update with the token information.

Still, I am curious as to why I can't replicate the code from the docs example. I will paste the code which worked for me below in hopes it will help.

```
# Client ID and Client Secret are set as env variables

import spotipy
from spotipy.oauth2 import SpotifyOAuth

scope = "user-top-read"
OAuth = SpotifyOAuth(scope=scope,
                     redirect_uri='http://localhost:8888',
                     cache_path='../cache.txt')
token = OAuth.get_access_token()

# receive the following warning
# __main__:1: DeprecationWarning: You're using 'as_dict = True'.get_access_token will return the token string directly in future 
#  versions. Please adjust your code accordingly, or use get_cached_token instead.
# At this point, I am taken to the user authorization and grant access with the 'user-top-read' scope

sp = spotipy.Spotify(auth_manager=OAuth)
top_tracks = sp.current_user_top_tracks()
```

stephanebruckert:
I agree that doesn't seem to behave correctly. I see at least two things that we need to fix:

 - we shouldn't have to specify a username and setting one should be optional. So `You must either set a cache_path or a username.` should be either removed or worse case be a warning
 - `SpotifyOAuth` is currently not completing the authentication. It just "sets" all the parameters needed and waits for an endpoint to be called. You are right, it should prompt user access immediately

a-camarillo:
[...]

Issue #512, opened 2020-06-13
Error 404


hafraz07:
```
import sys
import spotipy
import spotipy.util as util
from datetime import date

CLIENT_ID = ''
CLIENT_SECRET = ''
REDIRECT_URI = ''

def create_playlist(res, sp_obj, user):
    playlist_name = date.today().strftime('%m-%d')
    # Checks if playlist already exists
    playlists = sp_obj.user_playlists(user)
    for playlist in playlists['items']:
        if playlist['name'] == playlist_name:
            print('Playlist', playlist_name, 'already exists')
            return
    # Adds top tracks to track_ids
    track_ids = []
    for i, item in enumerate(res['items']):
        track_ids.append(item['track']['id'])

    # Creates playlist with name MM-DD
    sp_obj.user_playlist_create(user, playlist_name)

    # Gets playlist id of created playlist
    playlists = sp_obj.user_playlists(user)
    for playlist in playlists['items']:
        if playlist['name'] == playlist_name:
            playlist_id = playlist['id']
            break
    # Adds 50 recently played tracks to the playlist
    sp_obj.user_playlist_add_tracks(user, playlist_id, track_ids)
    print('Successfully created playlist', playlist_name)

if __name__ == '__main__':
    if len(sys.argv) > 1:
        username = sys.argv[1]
    else:
        print("Whoops, need your username!")
        print("usage: python playlist_generator.py [username]")
        sys.exit()

    scopes = 'playlist-modify-public user-top-read user-read-recently-played'

    token = util.prompt_for_user_token(username, scopes, CLIENT_ID, CLIENT_SECRET, REDIRECT_URI)

    if token:
        sp = spotipy.Spotify(auth=token)
        results = sp.current_user_recently_played(limit=50)
        create_playlist(results, sp, username)
    else:
        print("Can't get token for", username)
```
Hi, I made this script to create a playlist based on my recently played tracks. This works for my own account but currently gets a 404 when my friend tries to run the script with his account. I don't know if I'm doing something wrong or if this is an issue with the library. Thanks

stephanebruckert:
Hi @hafraz07, on which line does your friend get a 404? the stack trace would be helpful also. Thanks

hafraz07:
```
HTTP Error for GET to https://api.spotify.com/v1/users/farahzabinrahman/playlists returned 404 due to Not found.
Traceback (most recent call last):
[...]

PR #567, opened 2020-08-30
Made cache_path and username optional

stephanebruckert:
As suggested here https://github.com/plamere/spotipy/issues/533#issuecomment-682304023

The API itself does not require a username but Spotipy was still asking for it in order to know where to store the token. 

I think most devs want to get started quickly and will by default only use their main account, so storing that main account token in a cache called `.cache` should be fine.

Apps that specify a username should still work.

Fixes https://github.com/plamere/spotipy/issues/507, https://github.com/plamere/spotipy/issues/512, https://github.com/plamere/spotipy/issues/549

### DIFF
--- a/spotipy/oauth2.py
+++ b/spotipy/oauth2.py
@@ -852,9 +852,9 @@ class SpotifyPKCE(SpotifyAuthBase):
             raise SpotifyOauthError('error: {0}, error_descr: {1}'.format(error_payload['error'],
                                                                           error_payload[
                                                                               'error_description'
-                                                                              ]),
-                                    error=error_payload['error'],
-                                    error_description=error_payload['error_description'])
+            ]),
+                error=error_payload['error'],
+                error_description=error_payload['error_description'])
         token_info = response.json()
         token_info = self._add_custom_values_to_token_info(token_info)
         self._save_token_info(token_info)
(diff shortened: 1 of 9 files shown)


spotipy/oauth2.py
```diff
@@ -272,33 +286,26 @@ def get_cached_token(self):
...
+                token_info = self.refresh_access_token(
+                    token_info["refresh_token"]
+                )
 
-                if self.is_token_expired(token_info):
-                    token_info = self.refresh_access_token(
-                        token_info["refresh_token"]
-                    )
+        except IOError:
+            pass
```
DJSdev, line 307: I know you simply fixed up these try/catch statements. Could we throw in a logger warning similar to line 1026?
```suggestion
            logger.warning("Couldn't read cache at: %s", self.cache_path)
```

spotipy/oauth2.py
```diff
@@ -744,33 +756,26 @@ def get_cached_token(self):
...
```
DJSdev, line 777: Same as previous suggestion for logger warning

spotipy/oauth2.py
```diff
@@ -990,44 +997,33 @@ def get_cached_token(self):
...
-                    self.scope, token_info["scope"]
-                ):
-                    return None
+            if self.is_token_expired(token_info):
+                return None
 
-                if self.is_token_expired(token_info):
-                    return None
+        except IOError:
+            pass
```
DJSdev, line 1016: Same as previous suggestions for logger warning

spotipy/oauth2.py
stephanebruckert: Yep

G-S-Paris: I ended up doing something similar locally, I like the fix

stephanebruckert: Thanks, I just released it in 2.15.0

Issue #572, opened 2020-09-05
How to get an OAuth token on a headless machine?


RaddedMC:
Hey there! I'm not sure if this is the right place to write about my issue, but I'm working on adding spotify integration to my SmartFrame infoboard using the Authorization code flow. The code works perfectly running on my desktop (ubuntu 20.04 LTS), but on my raspberry pi (zero w running dietpi) that is running the infoboard client, the code hangs when calling spotify.current_user_playing_track()
I've dug around and discovered that the problem is related to my raspberry pi being a headless machine, so there is no way for me to log into spotify via a web browser like how I did on my desktop, meaning spotipy is just hanging waiting for an OAuth token.
How would I get an OAuth token on my headless Raspberry PI? I've heard that I can host a web server on the pi but I've never hosted my own web server so I'm not sure where to start there. Would it be possible to copy/paste the OAuth token from my desktop to my rpi? I know that breaks like a million best practices but this is just for a personal project running on my home network and not for anything that's going to be publicly released.
Thanks for any and all help?

trimmed down code, appropriate try/catches and data collection are in my final code:
```python
import spotipy

from spotipy.oauth2 import SpotifyOAuth

spotify_scope = "user-read-currently-playing"

spotify_user = #open a file and get my spotify username

spotify = spotipy.Spotify(auth_manager=SpotifyOAuth(scope=spotify_scope, redirect_uri="http://127.0.0.1:9090", 
cache_path=".spotifcache", client_id="MYID", client_secret="MYSECRET"))

current_track = spotify.current_user_playing_track() #This causes the code to hang

current_list =spotify.user_playlist(user=spotify_user, playlist_id=current_track["context"]["uri"], fields="name") #This never runs, trycatches are in place incase this returns nothing
```

stephanebruckert:
To solve this we need to do https://github.com/plamere/spotipy/issues/560

PR #574, opened 2020-09-11
Allow open_browser to be set on creation of auth_manager (SpotifyOAuth/SpotifyPKCE)

### LINKED ISSUE #560: Set open_browser on creation of auth_manager (SpotifyOAuth/SpotifyPKCE)
I use Spotipy in some scripts that I am running remotely ssh'd into servers where I don't have access to a browser. It looks like the auth flow is attempting to auto-open a browser, but nothing happens. My current workaround is to run the script and complete the auth flow locally and paste the resulting url + code back into the terminal.

Since I know ahead of time that I won't have access to a browser, and because there's already an `open_browser` variable passed around during the auth flow, I'd love to be able to set this up front when instantiating the auth_manager so that instead of it trying to open a browser and failing, it would just give me the auth url so I could open it myself.

Niquolas:
This PR fixes #560 (and possibly #572)

One can now instantiate SpotifyOAuth or SpotifyPKCE with `open_browser` set to `False`, behaving the same as if you had called `get_auth_response(open_browser=False)`. For backwards compatibility, this defaults to `True`.

Example:

```python
import spotipy

from spotipy.oauth2 import SpotifyOAuth

oauth = SpotifyOAuth(scope="streaming", open_browser=False)

spotify = spotipy.Spotify(auth_manager=oauth)

if __name__ == "__main__":
    spotify.start_playback()
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-// Add your changes here and then delete this line
+### Added
+
+- `open_browser` can be passed to the constructors of `SpotifyOAuth` and `SpotifyPKCE` to make it easier to authorize in browserless environments
 
 ## [2.15.0] - 2020-09-08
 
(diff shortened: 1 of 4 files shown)


Niquolas:
Just a thought, my solution prevents `get_auth_response(open_browser=False)` from working if it's set to `True` in the constructor. I'll look into that tomorrow.

stephanebruckert:
Nice, thanks for that. 

Suggestion: we could add your snippet as an example in /examples and also since many are looking for the headless mode, we might want to update the FAQ with a short how-to

stephanebruckert (approved):
:ok_hand:

stephanebruckert:
Looks great, merging and will create a release in a few days. Thanks!

PR #578, opened 2020-09-23
use a python3 approach please

tousif101:
My application is failing because of this


stephanebruckert:
Hey @tousif101, thanks for that. 

This would need to work on both python2 and python3. What do you think of the following solutions?

 - https://stackoverflow.com/a/7321970/1515819
 - https://stackoverflow.com/a/31687067/1515819 

please also make sure the checks are passing

stephanebruckert:
I think the idea makes sense but due to inactivity I'm closing it, feel free to reopen

Issue #581, opened 2020-10-05
Multiple errors thrown on reaching max retries by #580


Quizz1Cal:
*NOTE: My first issue to a public repo, any feedback is appreciated :)* 

**Describe the bug**
Using spotipy post-#580 (relevant issue #571 ) yields uncaught urllib3 / requests library errors, and in particular an AttributeError due to `retry_error.response` being None (see below). Furthermore, though not featured below, line 274 of client.py under this commit will also throw an AttributeError for the same reason.

**Your code**
*Note*: Have defined environment variables `SPOTIPY_CLIENT_ID`, `SPOTIPY_CLIENT_SECRET`, `SPOTIPY_CLIENT_USERNAME` and `SPOTIPY_REDIRECT_URI`.
```python
# GetClientCredentials() also works here, the playlist is public
spotify = Spotify(auth_manager=SpotifyOAuth(scope='playlist-read-private'))
brain_food_playlist_id = '37i9dQZF1DWXLeA8Omikj7'
playlist = spotify.playlist(playlist_id=brain_food_playlist_id)
print(spotify.playlist_is_following(playlist_id=playlist['id'], user_ids=['spotify:user:bad_username']))
```

**Expected behavior**
According to the implementation of the `requests` library, responses are expected to not be passed to RetryErrors ([see source code here](https://github.com/psf/requests/blob/master/requests/adapters.py#L507); AFAIK this is the only circumstance it is thrown by the library). 

Therefore, it is my belief that:
1. A SpotifyException should be thrown (potentially, with the associated url and header, if retrievable)
2. (if would not unreasonably impact existing behaviour) The non-spotipy exceptions should not be thrown

**Output**
**WARNING**: Paths modified for privacy
```
Max Retries reached
Traceback (most recent call last):
  File "../project/venv/lib/python3.8/site-packages/requests/adapters.py", line 439, in send
    resp = conn.urlopen(
  File "../project/venv/lib/python3.8/site-packages/urllib3/connectionpool.py", line 817, in urlopen
    return self.urlopen(
  File "../project/venv/lib/python3.8/site-packages/urllib3/connectionpool.py", line 817, in urlopen
    return self.urlopen(
  File "../project/venv/lib/python3.8/site-packages/urllib3/connectionpool.py", line 817, in urlopen
    return self.urlopen(
  File "../project/venv/lib/python3.8/site-packages/urllib3/connectionpool.py", line 807, in urlopen
[...]

PR #583, opened 2020-10-06
Fixed UnboundLocalError / AttributeError #571, #581

### LINKED ISSUE #571: UnboundLocalError: local variable 'response' referenced before assignment
**Describe the bug**
If something goes wrong in `self._session.request()` the variable `response` will never be set. Leading to a 
`UnboundLocalError: local variable 'response' referenced before assignment`

**Your code**
Sorry.
Error from my Rollbar:
Screen Shot 2020-09-05 at 9 51 41 AM

**Expected behavior**
Requests fail and that's fine. Judging from skimming the code, it's trying to package a message to go with raising a `SpotifyException` but it's failing to do so. 

**Output**
Paste and format errors (with complete stacktrace) or logs. Make sure to remove sensitive information.

**Environment:**
 - spotipy version: `spotipy==2.13.0`

**Additional context**
Relying on `UnboundLocalError` from a caller isn't nice. If it was `SpotifyException` I could do something like this:

```python
from spotipy.exceptions import SpotifyException

try:
    something_that_calls_spotipy(...)
except SpotifyException as err:
    logger.warning("Oh no. Spotify exception...")
```

Quizz1Cal:
Have also added an integration test to confirm that when max entries is reached, SpotifyException is thrown, and not the unexpected errors noted in issues #571 and #581.

Fixes #571 and #581.

### DIFF
--- a/tests/integration/test_non_user_endpoints.py
+++ b/tests/integration/test_non_user_endpoints.py
@@ -22,6 +22,7 @@ class AuthTestSpotipy(unittest.TestCase):
     """
 
     playlist = "spotify:user:plamere:playlist:2oCEWyyAPbZp9xhVSxZavx"
+    playlist_id = "2oCEWyyAPbZp9xhVSxZavx"
     four_tracks = ["spotify:track:6RtPijgfPKROxEzTHNRiDp",
                    "spotify:track:7IHOIqZUUInxjVkko181PB",
                    "4VrWlk8IQxevMvERoX08iC",
@@ -239,6 +240,13 @@ class AuthTestSpotipy(unittest.TestCase):
         # depending on the timing or bandwidth, this raises a timeout or connection error"
         self.assertRaises((requests.exceptions.Timeout, requests.exceptions.ConnectionError),
                           lambda: sp.search(q='my*', type='track'))
+
+    def test_max_retries_reached(self):
+        client_credentials_manager = SpotifyClientCredentials()
+        sp = spotipy.Spotify(client_credentials_manager=client_credentials_manager)
+        self.assertRaises((spotipy.exceptions.SpotifyException,),
+                          lambda: sp.playlist_is_following(playlist_id=AuthTestSpotipy.playlist_id,
+                                                           user_ids=[AuthTestSpotipy.bad_id]))
 
     def test_album_search(self):
         results = self.spotify.search(q='weezer pinkerton', type='album')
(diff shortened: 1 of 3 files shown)


tests/integration/test_non_user_endpoints.py
```diff
@@ -240,6 +241,13 @@ def test_search_timeout(self):
         self.assertRaises((requests.exceptions.Timeout, requests.exceptions.ConnectionError),
                           lambda: sp.search(q='my*', type='track'))
 
+    def test_max_retries_reached(self):
+        client_credentials_manager = SpotifyClientCredentials()
+        sp = spotipy.Spotify(client_credentials_manager=client_credentials_manager)
+        self.assertRaises((spotipy.exceptions.SpotifyException,),
+                          lambda: sp.playlist_is_following(playlist_id=AuthTestSpotipy.playlist_id,
+                                                           user_ids=[AuthTestSpotipy.bad_id]))
```
stephanebruckert, line 249: Your test doesn't seem to test the retry code. I had a look and came up with the following:


```suggestion
        spotify_no_retry = Spotify(
            client_credentials_manager=SpotifyClientCredentials(),
            retries=0)
        i = 0
        while i < 100:
            try:
                spotify_no_retry.search(q='foo')
            except spotipy.exceptions.SpotifyException as e:
                self.assertIsInstance(e, spotipy.exceptions.SpotifyException)
                return
            i += 1
        self.fail()
```

it's not ideal as it's doing many calls without mocking, but it should do for now

tests/integration/test_non_user_endpoints.py
```diff
@@ -22,6 +22,7 @@ class AuthTestSpotipy(unittest.TestCase):
     """
 
     playlist = "spotify:user:plamere:playlist:2oCEWyyAPbZp9xhVSxZavx"
+    playlist_id = "2oCEWyyAPbZp9xhVSxZavx"
```
stephanebruckert, line 25: to remove if you are going to use the suggestion below

stephanebruckert (commented):
Code looks good, I added a suggestion for the test

tests/integration/test_non_user_endpoints.py
Quizz1Cal: Fair point, I was going for the 'many-call' approach but in retrospect I agree a mock-based approach is better.

Quizz1Cal:
Changes agreed with and committed.

stephanebruckert:
Amazing, thanks for finding and fixing that one! 👌

PR #589, opened 2020-10-18
isort Module is Observed in /spotipy/*

lnxpy:
I installed the `isort` module right beside the other dependencies. Now, isort is observed in /spotipy/*.py files. There is no big change in the main algorithm.

> I'm contributing in Hacktoberfest challenge. Be sure that I'll make some essential changes in the project. I just need to add the "hacktoberfest" label in this PR.

Thanks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - SpotifyException now thrown when a request fails & has no response ( fixes #571, #581 )
 
-## [2.16.0] - 2020-09-18
-
 ### Changed
 
 - both inline and starting import lists are sorted using `isort` module
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,11 +5,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+
 ## Unreleased
 
 ### Fixed
 
 - SpotifyException now thrown when a request fails & has no response ( fixes #571, #581 )
+
+## [2.16.0] - 2020-09-18
+
+### Changed
+
+- both inline and starting import lists are sorted using `isort` module
 
 ## [2.16.0] - 2020-09-16
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,6 +31,11 @@ To verify the code style:
     pip install flake8
     flake8 .
 
+To make sure if the import lists are stored correctly:
+
+    pip install isort
+    isort . -c -v
+
 ### Changelog
 
 Don't forget to add a short description of your change in the [CHANGELOG](CHANGELOG.md)
--- a/spotipy/__init__.py
+++ b/spotipy/__init__.py
@@ -1,4 +1,4 @@
 from .client import *  # noqa
+from .exceptions import *  # noqa
 from .oauth2 import *  # noqa
 from .util import *  # noqa
-from .exceptions import *  # noqa
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -9,8 +9,8 @@ import logging
 import warnings
 
 import requests
-import urllib3
 import six
+import urllib3
 
 from spotipy.exceptions import SpotifyException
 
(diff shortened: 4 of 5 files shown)


stephanebruckert:
Thanks for that. Please update the changelog :-) 

Also, how does one use "isort"? is there a command we can use to sort imports in the future? could we add this as a [CI check](https://github.com/plamere/spotipy/blob/88ad7e05d515740844c7f654f81f4fddf3e517fa/.github/workflows/pythonapp.yml#L23)?

lnxpy:
Thanks for reposing soon. I'm working on the logging side of the Spotipy now. (using the default script-based configurations)
Be sure, I'll update the changelog file.

I do guess it brings some troubles in using isort module in the CI services. I've worked on a lot of projects and used isort in order to sort the import lists but in some cases, it just failed in tests!! (e.g. you can take a look at this closed [PR on IBM/lale](https://github.com/IBM/lale/pull/473))

CHANGELOG.md
```diff
@@ -5,12 +5,19 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+
```
stephanebruckert, line 8: extra line to remove

CHANGELOG.md
```diff
@@ -5,12 +5,19 @@ All notable changes to this project will be documented in this file.
...
 ## Unreleased
 
 ### Fixed
 
 - SpotifyException now thrown when a request fails & has no response ( fixes #571, #581 )
 
+## [2.16.0] - 2020-09-18
```
stephanebruckert, line 15: You can remove this line as 2.16.0 was already released and your changes are still unreleased

stephanebruckert (changes requested):
We could merge your PR as it is, but it would be better if others can profit from it:

 1. what command did you run? add it to https://github.com/plamere/spotipy/blob/master/CONTRIBUTING.md#lint
 2. in our case we could probably run a command that checks if imports are sorted or not, see https://github.com/PyCQA/isort#the---check-only-option. So maybe, after this line

      https://github.com/plamere/spotipy/blob/88ad7e05d515740844c7f654f81f4fddf3e517fa/.github/workflows/pythonapp.yml#L27

     add the command you ran in (1) by adding `-c` to it

  3. As shown in https://github.com/PyCQA/isort#setuptools-integration, I'm thinking we should add `isort` to the spotipy `extra_reqs`: https://github.com/plamere/spotipy/blob/88ad7e05d515740844c7f654f81f4fddf3e517fa/setup.py#L14

lnxpy (commented):
Everything is ok with these changes. Flake has restored some changes in the scripts which are completely compatible in isort.

lnxpy:
I suggest you add the flake8 plugin instead of using isort. It helps you to verify if all import lists are sorted. [Check flake8-isort module](https://pypi.org/project/flake8-isort/). (I've never worked with this tool and that why I have no idea about editing the lint section so I leave it on your own)

Thank you so much. :)

stephanebruckert:
> I've never worked with this tool and that why I have no idea about editing the lint section so I leave it on your own

You have to start sometime @lnxpy 😺 

Your PR in its current state does not break anything so that's great, but it also doesn't add long-term value. Codebase changes should come with tests and changing the style of a codebase without adding an automatic rule is useful only temporarily because others won't follow it and soon, your initial sorting will be erased. You probably wouldn't like to open that same PR in a month, would you?

For your PR to get the value it deserves and be merged, it would need:
 - instructions for other contributors to do the same, so they don't have to look for it and we won't loose your initial investment (please don't do this),
 - even better, an automatic rule so that we don't need to worry about it ever again (please do this).

> I suggest you add the flake8 plugin instead of using isort

I also agree that using the flake8-isort plugin is better than using isort itself, because we already use flake8. Please note that there is also https://pypi.org/project/flake8-import-order/, feel free to pick your favorite

lnxpy:
Thanks @stephanebruckert,
I'm kinda confused now. :sweat_smile: Where should I sign? Is there any other file to change and submit these changes? (Because I've already changed the changelog file)

stephanebruckert:
@lnxpy that's fine, what command did you run to sort the imports? please add that command to the following doc, so that next contributors know how to do it: https://github.com/plamere/spotipy/blob/master/CONTRIBUTING.md#lint and we can merge your PR

lnxpy:
I've already worked on the logging system. I suppose to make another PR for those changes.
Thanks again. Such a long conversation!! :D

stephanebruckert:
Yeah I just wanted to express how important it is to share your knowledge, which you did, so thank you. Looking forward to your next PR!

PR #590, opened 2020-10-20
Default Script-based Configurations is set for Logging

lnxpy:
In most cases, It's better to create a base configuration for `logging` which is easily available on [structlog](https://www.structlog.org/en/stable/standard-library.html) too. Not a big difference to either use a YML, or a JSON based method for configurations. Most often we use the external methods when we are trying to focus on the logging side. Base configurations allow you to override the original classes and have your own configurations.

#### Changes
- `CHANGELOG.md` file ~ "Unreleased" section: Subsections are sorted alphabetically and the new changes and adds are explained.
- `CONTRIBUTING.md` file ~ "Create virtual environment" section: When you create a virtual environment, there is no need to use `--user` flag in the installation process. In some cases, it might break the installation cycle. Packages are being installed in the virtual environment by default, but if you use `--user`, it will force it to install outside the virtual environments.

> I enthusiastically appreciate it if you add the `hacktoberfest` label to this PR too. Thank you.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,13 +8,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-### Fixed
+### Added
 
-- SpotifyException now thrown when a request fails & has no response ( fixes #571, #581 )
+- Base script-based configurations observed for `logging` side
 
 ### Changed
 
 - both inline and starting import lists are sorted using `isort` module
+
+### Fixed
+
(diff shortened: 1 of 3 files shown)


stephanebruckert:
Thanks for this, however I don't think the library itself should initialize `basicConfig`, see that this is something we have reverted before https://github.com/plamere/spotipy/pull/482. My understanding is that it should be up to the developer to choose how to initialise it in their own apps.

We will likely close this, unless someone has a better idea?

PR #597, opened 2020-10-24
local language and stop

busterbeam:
- Added defaulting to local language on computer which script is run on.
- Stop playback function on track and context level.
- Literal string interpolation change from old string formatting, for better readability and compactness.
- Separated string join lists from functions for further compactness and reduction on repeated code


busterbeam:
All builds are failing as they do not understand literal string interpolation, `trackid` is used in literal string but is not recognized by flake8.  Also python 2.7 does not support literal string interpolation.  I wouldn't think you supported python 2.7 still?  Sorry I didn't change the changelog my bad

PR #603, opened 2020-11-07
Small adjustment on sign_out function

migrmrz:
First, thanks for this. It has been of great help. I was having trouble with this function, getting a TypeError (TypeError: can only concatenate str (not "NoneType") to str) because apparently the second os.remove(session_cache_path()) is trying to delete a file that does not exist anymore so this is what worked for me.

### DIFF
--- a/examples/app.py
+++ b/examples/app.py
@@ -73,11 +73,10 @@ def index():
 
 @app.route('/sign_out')
 def sign_out():
-    os.remove(session_cache_path())
-    session.clear()
     try:
         # Remove the CACHE file (.cache-test) so that a new user can authorize.
         os.remove(session_cache_path())
+	session.clear()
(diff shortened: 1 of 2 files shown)


examples/app.py
```diff
@@ -73,11 +73,10 @@ def index():
 
 @app.route('/sign_out')
 def sign_out():
-    os.remove(session_cache_path())
-    session.clear()
     try:
         # Remove the CACHE file (.cache-test) so that a new user can authorize.
         os.remove(session_cache_path())
+	session.clear()
```
stephanebruckert, line 79: please fix indent

migrmrz: Sorry about that, it's corrected now.

stephanebruckert: thanks!

Issue #605, opened 2020-11-08
Spotify OAuth error : Couldn't read cache, Couldn't refresh token and Illegal URI 


siddharthnahar:
I have a couple of things that are going wrong, and I have spent a considerable amount of time to fix it. I was first able to run the code below, Initially I did not pass the cache path as a parameter to the auth_manager. The code worked for an hour (since the token expires every hour), I then started running into errors regarding token refresh. 

I then added the cache path as an added parameter and am getting two errors: 
1. Couldn't read cache at <cache_path>
2. Illegal URI <Enter the URL you were redirected to>. 

I have already checked the Redirect URI on my spotify developer dashboard and it matches what I have passed to the script. 

Could I get a way to fix this as well as refresh the token if it's expired? 

**code**

```
import configparser
import spotipy
from spotipy.oauth2 import SpotifyOAuth
import json
import os
import csv



class UserData():

    def __init__(self, config_path):
        config = configparser.ConfigParser()
        config.read(config_path)

        self.CLIENT_ID = config['AUTH']['cid']
        self.CLIENT_SECRET = config['AUTH']['secret']
        self.REDIRECT_URI = config['AUTH']['redirect_uri']
        self.USERNAME = config['AUTH']['username']
        self.CACHE = config['AUTH']['cache']
        self.HOME_DIR = config['PATH']['home_dir']
        self.top_artists = {}
        self.top_tracks = {}


    def get_top_artist(self, limit=None, time_range=None, offset=0):
        sp = spotipy.Spotify(auth_manager=SpotifyOAuth(client_id=self.CLIENT_ID,
                                                client_secret=self.CLIENT_SECRET,
                                                redirect_uri=self.REDIRECT_URI,
                                                scope="user-top-read", 
                                                username=self.USERNAME, 
                                                cache_path=self.CACHE))

        
        self.top_artists = sp.current_user_top_artists(limit=limit, offset=offset, time_range=time_range)

        with open(os.path.join(os.path.dirname(os.path.realpath(__file__)))+'/top_artist.csv', 'w' ,newline='') as csvfile:

            writer = csv.writer(csvfile, delimiter = ',', quoting = csv.QUOTE_MINIMAL)
            writer.writerow(['sno', 'id', 'name', 'popularity', 'genre', 'uri', 'image_ref', 'follower_count'])

            for idx, item in enumerate(self.top_artists['items']):
                sno = idx
                id = item['id']
                name = item['name']
                popularity = item['popularity']
[...]

Issue #599, opened 2020-10-28
"Invalid access token, reason: None" with access token from OAuth 


alansberman:
I'm new to both Django and spotipy, and am trying to get my currently playing track.

This is what I have thus far:

```
def home(request):
	sp = spotipy.Spotify(auth_manager=SpotifyOAuth(client_id="<my_token>",
                                               client_secret="<my_secret>",
                                               redirect_uri="http://localhost:8000/callback/",
                                               scope="user-library-read user-read-currently-playing"))
	results = sp.search(q='weezer', limit=20)
	return render(request, 'overview/dashboard.html')

def overview(request):
	code = request.GET.urlencode().split("=")[1]
	print(code, 'is code') # code (token) is returned
	if code:
		songs = []
		sp = spotipy.Spotify(auth=code)
		results = sp.current_user_saved_tracks()
		print(results)
		for idx, item in enumerate(results['items']):
			track = item['track']
			songs.append(track['artists'][0]['name'])
			return render(request, 'overview/dashboard.html', {'songs': songs})
	else:
		return 200
```

it appears I have to include the line ```results = sp.search(q='weezer', limit=20)``` or nothing happens.

Anyway, when I click allow on Spotify's popup and get to the callback, with a token, I get the following error 

```
SpotifyException at /callback/
http status: 401, code:-1 -https://api.spotify.com/v1/me/tracks?limit=20&offset=0:
 Invalid access token, reason: None
```

What exactly is going wrong here? Any help would be appreciated, and apologies for the ugly code!

Peter-Schorn:
The variable `code` in `def overview(request)` is probably the authorization code, as described in [the authorization guide][1], not the access token, which is what the `auth` parameter of `Spotify` expects. The Spotify web API does not return access tokens in query strings.

More importantly, though, why are you trying to implement the authorization process yourself? spotipy already has authorization managers that do that for you: `SpotifyClientCredentials`, `SpotifyOAuth`, `SpotifyPKCE`, and `SpotifyImplicitGrant`. You're making things more complicated than they need to be. All you need to do is the following:

```
from spotipy import Spotify, SpotifyOAuth

spotify = Spotify(
    auth_manager=SpotifyOAuth(
        client_id="client_id",
        client_secret="client_secret",
        redirect_uri="http://localhost:8080/callback/",
        scope="user-library-read user-read-currently-playing"
    )
)

results = spotify.current_user_saved_tracks()
print(results)
songs = []
[...]

PR #608, opened 2020-11-11
Suppress warnings when cache path does not exist.

Peter-Schorn:
See issue #605 

I also changed the docs for `auth` parameter of `Spotify.init` to "access token" instead of "authorization token". In issue #599, a user confused the access token with the authorization code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-// Add your changes here and then delete this line
+- A warning will no longer be emitted when the cache file does not exist at the specified path
+- The docs for the `auth` parameter of `Spotify.init` use the term "access token" instead of "authorization token"  
 
 ## [2.16.1] - 2020-10-24
 
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -114,7 +114,7 @@ class Spotify(object):
         """
         Creates a Spotify API client.
 
-        :param auth: An authorization token (optional)
+        :param auth: An access token (optional)
         :param requests_session:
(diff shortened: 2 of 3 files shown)


Peter-Schorn:
It looks like the unit tests are failing because they expect `open` to be called with the cache path. You might have to help me fix the tests.

stephanebruckert:
@Peter-Schorn to be honest I'm struggling with the tests as well, they would need a proper clean up.

Can I suggest we do it more simply: keep it like before but catch `FileNotFoundError`?

```
try:
    ..
except FileNotFoundError:
    ..
except IOError:
    ..
```

Peter-Schorn:
@stephanebruckert I would've preferred to use `FileNotFoundError` as well, but it doesn't exist in python 2.7. That's why I had to change it to `os.path.exists`.

Peter-Schorn:
I've been looking at the unit tests trying to figure out how they work, but I'm very confused. Is a cache file ever created for the tests?

stephanebruckert:
It seems all mocked to me, checking that the right functions are called, and forcing their return values. It looks like we force the token to be created from a JSON. 

These tests might not be perfect or very clear, but I trust they make some sense. I would really avoid refactoring them in this PR. I'm having a look at the methods here to see if we could catch missing files in 2.7 https://stackoverflow.com/q/21367320/1515819

stephanebruckert:
This would work in 2.7:

```python
import errno

try:
    open('./file')
except EnvironmentError as e:
    if e.errno == errno.ENOENT:
        print("File not found (file does not exist at all)")
    else:
        print("Can't read file (file exists, but might not have permission to read it")
```

stephanebruckert:
Nice one! Looks great to me, I can merge if you want?

Peter-Schorn:
Yes, go ahead.

PR #609, opened 2020-11-14
Added more logging and fixed docs for search endpoint

Peter-Schorn:
Logging the requests for retrieving the tokens and refreshing the tokens will make it easier to debug issues that users have with the library. For example, we could've solved #605 faster if these log messages were emitted. I also fixed the docs for the search endpoint to mention that you *can* search for multiple types at the same time.


stephanebruckert:
👍

PR #607, opened 2020-11-10
Add market argument to track info 

z1lc:
So that we know if a track is available to be played in a specific country.

I've been using this successfully in production (through a forked version) at https://github.com/z1lc/zdone/blob/e5d56f32b17b45975a64e092f4449c20197c2015/app/spotify.py#L351.

PR #610, opened 2020-11-15
Add market argument to track info

z1lc:
New PR instead of https://github.com/plamere/spotipy/pull/607

So that we know if a track is available to be played in a specific country.

I've been using this successfully in production (through a forked version) at https://github.com/z1lc/zdone/blob/e5d56f32b17b45975a64e092f4449c20197c2015/app/spotify.py#L351.


stephanebruckert:
Thanks! I should be able to bump the version soon

PR #624, opened 2020-12-13
Fix examples for saved tracks

### LINKED ISSUE #622: delete_a_saved_track example fails
Hello

When executing the script, the track is not deleted from the saved tracks

https://github.com/plamere/spotipy/blob/2.16.1/examples/delete_a_saved_track.py

no error is output -> Process finished with exit code 0

Environment:
Python version [3.9.0]
spotipy version [2.16.1]

JoshuaBaunach:
This PR fixes #622 by fixing the indentation of the last three lines and fixing the way user input is handled.

There are similar errors in contains_a_saved_track.py. This PR fixes those as well.

### DIFF
--- a/examples/contains_a_saved_track.py
+++ b/examples/contains_a_saved_track.py
@@ -1,3 +1,5 @@
+# Prints whether a track exists in your collection of saved tracks
+
 import pprint
 import sys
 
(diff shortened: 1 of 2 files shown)


stephanebruckert:
Thanks for this. 

Don't worry about updating the changelog, no need for examples

Merging!

PR #625, opened 2020-12-16
Create CacheHandler to abstract caching tokens

tonyaajjackson:
## Contributing.md Checklist:

- [X] Tests pass
- [X] Autopep8 fixing performed
- [X] flake8 verification passes
- [ ] isort: raises issues, but issues appear unrelated to changes in pull request

## Change Description
Previous code only supported caching to and from json files in a given directory. In addition, the get_cached_token method mixed getting and getting the token in the same method.

This change creates a `CacheHandler` class to abstract out the caching implementation and allow the user to cache tokens in any way they see fit. For example, the user could create a `MongoCache` class to store and retrieve tokens from a Mongo database and specify that `cache_handler=MongoCache` when creating an `auth_manager` object.

`cache_handler` is added as an argument to `SpotifyOAuth`, `SpotifyPKCE`, and `SpotifyImplicitGrant`. Specifying a `cache_handler` now overrides any specification of `cache_path` and/or `username`.

To preserve backwards compatibility in handling cache files, a `CacheFileHandler` class extending `CacheHandler` is created that implements previous functionality in getting and saving to json files. If no `cache_handler` is specified, the `cache_path` and `username` arguments are used to create an instance of `CacheFileHandler`. It may be worth deprecating the `cache_path` and `username` fields in favour of using `CacheFileHandler` to simplify the interface.

Tests are also modified and extended to cover the new functionality. A sample `MemoryCache` `CacheHandler` that stores the cached token as a variable in memory is created to test getting and saving to a custom `CacheHandler`.

Minor refactoring of `is_token_expired` and `_is_scope_subset` also performed to make adding the `CacheHandler` abstraction easier.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added log messages for when the access and refresh tokens are retrieved and when they are refreshed
-- Support `market` optional parameter in `track`  
+- Support `market` optional parameter in `track`
+- Added CacheHandler abstraction to allow users to cache tokens in any way they see fit
 
 ## [2.16.1] - 2020-10-24
 
--- a/spotipy/__init__.py
+++ b/spotipy/__init__.py
@@ -1,3 +1,4 @@
+from .cache_handler import * #noqa
 from .client import *  # noqa
 from .exceptions import *  # noqa
 from .oauth2 import *  # noqa
--- a/spotipy/__init__.py
+++ b/spotipy/__init__.py
@@ -1,4 +1,4 @@
-from .cache_handler import * #noqa
+from .cache_handler import *  # noqa
 from .client import *  # noqa
 from .exceptions import *  # noqa
 from .oauth2 import *  # noqa
--- a/spotipy/cache_handler.py
+++ b/spotipy/cache_handler.py
@@ -1,0 +1,82 @@
+import errno
+import json
+import logging
+
+logger = logging.getLogger(__name__)
+
+class CacheHandler():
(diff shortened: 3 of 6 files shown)


tonyaajjackson:
Looks like tests are failing on python 2.7 - I'll fix that tomorrow.

stephanebruckert:
That's amazing, thanks.

Since `oauth.py `is very fat, we could move `CacheHandler` to a new file, perhaps `cache.py` or `cache_handler.py`. Then we could have another folder named `cache_handlers` that would contain all the handlers we have. So far we have just the default one, `CacheFileHandler`, but later people could contribute their own

tonyaajjackson:
Good idea, Stephane! I've split the CacheHandler code into cache_handler.py and fixed the failing tests. Please let me know if there's any additional content you would like added to this pull request to make it ready for merging into the main repo.

Also, what are your thoughts on adding a deprecation warning on the `cache_path` and `username` arguments?

stephanebruckert:
Thanks for the fixes, going to merge it. 

I have a handler for DynamoDB in mind, which I hope to push soon

> Also, what are your thoughts on adding a deprecation warning on the cache_path and username arguments?

https://github.com/plamere/spotipy/blob/master/examples/app.py

 - `cache_path` is still necessary in the above example, a back-end that can handle multiple users. If we were to deprecate this field, we would need another way of achieving that 
 - I'm not so sure if `username` is still really needed, but for now it seems to  provide another way of setting the cache path

Happy to  discuss this in more detail in a fresh issue / PR though

tonyaajjackson:
Discussion of deprecation is in https://github.com/plamere/spotipy/pull/630

PR #627, opened 2020-12-19
Sorry 

rohan-ynwa:
I apologize, I accidentally requested a change and have no idea how to revese it


ritiek:
@rohan-ynwa You should revoke your current credentials ASAP (client id and client secret) on your Spotify developer app, since you published them in the open here. Except that it's not much of a big deal. Anything specific you were trying to do?

PR #628, opened 2020-12-25
Fix Spotify.user_playlist_reorder_items parameter order

pmdevita:
Spotify.user_playlist_reorder_items currently calls Spotify.playlist_reorder_items with the parameters in the incorrect order. Not a big deal since this is deprecated but nice for legacy projects.


stephanebruckert (approved):
Nice catch

PR #630, opened 2020-12-30
Add deprecation warnings to direct users towards using cache_handler

tonyaajjackson:
When `cache_path` or `username` are specified in the constructors of the `SpotifyOAuth`, `SpotifyPKCE`, or `SpotifyImplicitGrant` auth_managers, the constructor creates a `CacheFileHandler` instance under the hood and sets it as the auth_manager's `cache_handler`. The user is currently able to create a `CacheFileHandler` instance in two ways:

1. By creating it outside the auth_manager constructor and passing it as the `cache_handler` (new approach)
1. By passing `cache_path` and `username` to the auth_manager constructor (old approach)

Ideally, there would be one and only one obvious way to specify a CacheFileHandler instance. Specifying by passing to the `cache_handler` allows flexibility in which `CacheHandler` is used and should be preferred, so passing the `cache_path` or `username` to the auth_manager constructor should be deprecated.

This pull request adds deprecation warnings for if `cache_handler` or `username` are passed as arguments to `SpotifyOAuth`, `SpotifyPKCE`, or `SpotifyImplicitGrant`. In addition, the flask example is updated to use the `cache_handler` approach. I did not find other examples that used the `cache_path` or `username` arguments to the auth_manager constructors.

In addition, in https://github.com/plamere/spotipy/commit/9550c8fd86314a15d693a3389b68468b17813156 I accidentally broke the caching functionality in the `SpotifyOAUth`, `SpotifyPKCE`, and `SpotifyImplicitGrant` auth_managers by removing the `get_cached_token` and `_save_token_info` methods from the auth_manager classes. Users with existing codebases that use the `get_cached_token` and `_save_token_info` methods directly will experience errors if they upgrade spotipy.

This pull request restores the `get_cached_token` and `_save_token_info` methods on the three auth_manager classes as aliases for the corresponding methods in the `cache_handler`. Deprecation warnings are also added to the `get_cached_token` and `_save_token_info` methods to direct users to switch to using the new `cache_handler` approach. "Legacy" tests checking the `get_cached_token` and `_save_token_info` methods in the auth_managers are also added to ensure that the old functionality is preserved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The docs for the `auth` parameter of `Spotify.init` use the term "access token" instead of "authorization token"
 - Changed docs for `search` to mention that you can provide multiple types to search for
 - The query parameters of requests are now logged
+- Deprecate specifing `cache_path` or `username` directly to `SpotifyOAuth`, `SpotifyPKCE`, and `SpotifyImplicitGrant` constructors, instead directing users to use the `CacheFileHandler` cache handler
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added log messages for when the access and refresh tokens are retrieved and when they are refreshed
-- Support `market` optional parameter in `track`  
+- Support `market` optional parameter in `track`
+- Added CacheHandler abstraction to allow users to cache tokens in any way they see fit
 
 ## [2.16.1] - 2020-10-24
 
--- a/examples/app.py
+++ b/examples/app.py
@@ -48,8 +48,9 @@ def index():
         # Step 1. Visitor is unknown, give random ID
         session['uuid'] = str(uuid.uuid4())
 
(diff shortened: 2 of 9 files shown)


stephanebruckert (commented):
https://github.com/plamere/spotipy/blob/ba9ad070e9973ac2be0fe4842a6e538b14616789/spotipy/oauth2.py#L263

https://github.com/plamere/spotipy/blob/ba9ad070e9973ac2be0fe4842a6e538b14616789/spotipy/oauth2.py#L265

Would be good to mention deprecation in the parameter info for all `username` and `cache_path`  constructor fields.

Do you also know what takes precedence if someone accidentally specifies `username` both in a `SpotifyAuthBase` and `CacheFileHandler `? We could mention this in the docstring too

Otherwise looks great! thanks for that

tonyaajjackson:
I updated the docstrings to indicate that those parameters are deprecated - good catch.

If a `cache_handler` is specified, it takes precedence over both the `cache_path` and `username` fields, and the `cache_path` and `username` fields are discarded. I've added a warning for if the user specifies a `cache_handler` as well as a `cache_path` or `username`. In addition, I've rearranged the warning logic so that a deprecation warning is always triggered if `cache_path` or `username` are specified in the `SpotifyOAuth` & etc constructors. Previously, specifying a `cache_handler` would prevent the deprecation warning from being raised.

stephanebruckert (approved):

PR #636, opened 2021-01-30
Removed requirement for examples/app.py to specify port multiple times (only SPOTIPY_REDIRECT_URI needs to contain the port)

KnallbertLp:
we try to get the port from SPOTIPY_REDIRECT_URI as we already have to set it there anyway.
if we don't succeed we use 8080 as fallback.
for legacy compatibility we still allow PORT envar and let it override the setting

### DIFF
--- a/examples/app.py
+++ b/examples/app.py
@@ -124,5 +124,5 @@ Following lines allow application to be run more conveniently with
 (Also includes directive to leverage pythons threading capacity.)
 '''
 if __name__ == '__main__':
-	    app.run(threaded=True, port=int(os.environ.get("PORT", (re.findall("^\d+$|(?<=:)\d+$", os.environ.get("SPOTIPY_REDIRECT_URI", 8080)))[0])))
+    app.run(threaded=True, port=int(os.environ.get("PORT", (re.findall("^\d+$|(?<=:)\d+$", os.environ.get("SPOTIPY_REDIRECT_URI", 8080)))[0])))
 
--- a/examples/app.py
+++ b/examples/app.py
@@ -124,5 +124,7 @@ Following lines allow application to be run more conveniently with
 (Also includes directive to leverage pythons threading capacity.)
 '''
 if __name__ == '__main__':
-    app.run(threaded=True, port=int(os.environ.get("PORT", (re.findall("^\d+$|(?<=:)\d+$", os.environ.get("SPOTIPY_REDIRECT_URI", 8080)))[0])))
-
+    app.run(threaded=True, port=int(os.environ.get("PORT", int(
+        os.environ.get("SPOTIPY_REDIRECT_URI").split(":")[-1]) if os.environ.get("SPOTIPY_REDIRECT_URI") is not None and
+                                                                  os.environ.get("SPOTIPY_REDIRECT_URI").split(":")[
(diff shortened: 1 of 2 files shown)


examples/app.py
```diff
@@ -123,4 +124,5 @@ def current_user():
 (Also includes directive to leverage pythons threading capacity.)
 '''
 if __name__ == '__main__':
-	app.run(threaded=True, port=int(os.environ.get("PORT", 8080)))
+	    app.run(threaded=True, port=int(os.environ.get("PORT", (re.findall("^\d+$|(?<=:)\d+$", os.environ.get("SPOTIPY_REDIRECT_URI", 8080)))[0])))
```
stephanebruckert, line 127: Not a fan of the difficult regex here. To get what follows the last colon, could we not do:

    redirect_uri.split(":")[-1]

stephanebruckert (changes requested):
Good idea! Requested one change to make the code simpler

examples/app.py
stephanebruckert: also indentation needs fixing

KnallbertLp: probably, but the regex not only matches the port in SPOTIPY_REDIRECT_URI but also the default 8080
I don't know of the top of my head how we could do this as a oneliner.

KnallbertLp: If we do it with more than one line, the best readable version probably would be 
`    port = 8080`
`    port = int(os.environ.get("SPOTIPY_REDIRECT_URI").split(":")[-1]) if os.environ.get("PORT").split(":")[-1] else port`
`    port = int(os.environ.get("PORT")) if os.environ.get("PORT") else port`
`    app.run(threaded=True, port=port)`

KnallbertLp: never mind, I figured out a way. 
But still, it does not look nice, but at least is more readable then the ont-the-fly fix I previously stated

stephanebruckert: There is no need to specifically catch numbers, since we expect something that looks like DOMAIN:PORT

```python
app.run(threaded=True, port=int(os.environ.get("PORT",
                                os.environ.get("SPOTIPY_REDIRECT_URI", 8080).split(":")[-1])))
```

This defaults to 8080, is still a one-liner, is shorter and simpler to understand.

examples/app.py
```diff
@@ -123,4 +123,7 @@ def current_user():
 (Also includes directive to leverage pythons threading capacity.)
 '''
 if __name__ == '__main__':
-	app.run(threaded=True, port=int(os.environ.get("PORT", 8080)))
+    app.run(threaded=True, port=int(os.environ.get("PORT", int(
+        os.environ.get("SPOTIPY_REDIRECT_URI").split(":")[-1]) if os.environ.get("SPOTIPY_REDIRECT_URI") is not None and
+                                                                  os.environ.get("SPOTIPY_REDIRECT_URI").split(":")[
+                                                                      -1] is not None else 8080)))
```
stephanebruckert, line 129: have a look at my other comment, this can be simplified

examples/app.py
KnallbertLp: That looks very nice, thanks!

stephanebruckert:
Thanks!

Issue #637, opened 2021-02-06
Urllib3 method_whitelist deprecated


ReneNulschDE:
**Describe the bug**
In Urllib3 version >= 1.26.0 `Retry(method_whitelist=...)` is deprecated. Your component is using method_whitelist [here](https://github.com/plamere/spotipy/blob/44970c3348f4242be4dbadadf1d08ff5242918da/spotipy/client.py#L197)

**Your code**
None. Your lib is used in the Home-Assistant Spotify component.

**Expected behavior**
No deprecation warnings. Check if a replacement with `allowed_methods` is possible. Tested this with a local code change and works good.

**Output**
/home/rene/repos/core/venv/lib/python3.8/site-packages/urllib3/util/retry.py:252: DeprecationWarning: Using 'method_whitelist' with Retry is deprecated and will be removed in v2.0. Use 'allowed_methods' instead

**Environment:**
 - OS: Linux Ubuntu 20.04
 - Python version: 3.8.6
 - spotipy version: 2.16.1

**Additional context**

stephanebruckert:
Released in 2.17.1

PR #638, opened 2021-02-06
Replace Urllib3 method_whitelist with allowed_methods

ReneNulschDE:
This is a proposed fix for #637 

Since Urllib3 v1.26.0 method_whitelist is deprecated and replaced with allowed_methods.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fixed Spotify.user_playlist_reorder_tracks calling Spotify.playlist_reorder_tracks with an incorrect parameter order
+- Fixed deprecated Urllib3 `Retry(method_whitelist=...)` in favor of `Retry(allowed_methods=...)`
 
 ## [2.16.1] - 2020-10-24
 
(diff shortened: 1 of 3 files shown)


stephanebruckert (approved):

stephanebruckert (changes requested):
Just saw that we allow requests 2.20.0
https://github.com/plamere/spotipy/blob/2.16.1/setup.py#L29

and that the minimum urllib3 requests 2.20.0 needs is <1.25 as can be seen here https://github.com/psf/requests/blob/v2.20.0/setup.py#L47

Therefore the change would not currently work. We would need to see if our code allows us to bump the requests version. But that might be difficult as we need to keep compatibility with python2. If you manage to upgrade requests that would be awesome 👍

ReneNulschDE:
Checked the same topic, let me try to dive deeper

ReneNulschDE:
Hi @stephanebruckert ,

I made some test with Requests 2.25 and six 1.15 under python 2.7.18 and at least the simple code examples running fine.

stephanebruckert:
@ReneNulschDE I will check ASAP, I need to run the integration tests manually

stephanebruckert:
Thanks for that @ReneNulschDE, merged & released!

PR #646, opened 2021-02-26
Update search query urls in client.py

kiselz:
[https://developer.spotify.com/documentation/web-api/reference/search/search/](https://developer.spotify.com/documentation/web-api/reference/search/search/) - This url does not seem to be valid.
I guess [https://developer.spotify.com/documentation/web-api/reference/search/](https://developer.spotify.com/documentation/web-api/reference/search/) is right link.


stephanebruckert:
Thank you!

PR #651, opened 2021-03-04
Make return structure for "Get Multiple..." type APIs consistent

badfarmerjohn:
Get-Multiple APIs have a JSON response structure of with a single entry whose value is a list containing the actual requested items as dictionaries. The audio_features function in client.py was adapted to unwrap the single value in commit 9c0b872e86db479b1ba38a826a7f1f77fa3d68b8, but other functions for similar APIs (eg. tracks, artists, albums, ...) have not been adapted, leading to an inconsistent structure in the returned dictionary. This change modifies those functions to handle the response similar to audio_features.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,21 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Changed
+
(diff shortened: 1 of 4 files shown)


stephanebruckert:
Thanks for your PR, however I don't think we want this:

 - https://github.com/plamere/spotipy/commit/9c0b872e86db479b1ba38a826a7f1f77fa3d68b8 looks wrong to me, as according to [the doc of that endpoint](https://developer.spotify.com/documentation/web-api/reference/#endpoint-get-several-audio-features) we expect `audio_features`, not `audio_attributes`. 
 - the above not being useful, it would be wrong to adapt other endpoints to it
 - the change would be backward-incompatible, as current apps are expecting such intermediary fields. You could still raise this in the discussion about spotipy3 https://github.com/plamere/spotipy/issues/652 so we can discuss it, but keep in mind that spotipy's philosophy is to remain "light" and to avoid modifying the resulting JSON

Issue #645, opened 2021-02-25
How to retrieve original response body in case of 400 errors.


shivam-kapila:
Is there a way to retrieve the original response body in case of 400 errors. e.g. The error message in case of error #503 is
```
{
    "error": "invalid_grant",
    "error_description": "Refresh token revoked"
}
```
But spotipy only shows 400 Bad Request and not the original message `Refresh token revoked.` A way to access the original JSON would help better debug issues in client apps.

Peter-Schorn:
I just made a pull request that fixed your issue.

shivam-kapila:
@Peter-Schorn thanks for the action :)

stephanebruckert:
Released in 2.18.0

PR #655, opened 2021-03-12
Added cache handler to `SpotifyClientCredentials` and fixed a bug in refresh tokens methods that raised the wrong exception

Peter-Schorn:
I see no reason why we shouldn't save the token info for the Client Credentials flow to persistent storage as well.

The methods for refreshing the access token now raise the proper exception . For example, instead of
```
spotipy.exceptions.SpotifyException: http status: 400, code:-1 - Couldn't refresh token: code:400 reason:Bad Request, reason: {'Authorization': 'Basic ZDEz...ZDc='}
```
you get
```
spotipy.oauth2.SpotifyOauthError: error: invalid_grant, error_description: Refresh token revoked
```
See #645

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,14 +5,26 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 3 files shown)


stephanebruckert (approved):
> I see no reason why we shouldn't save the token info for the Client Credentials flow to persistent storage as well.

agreed!

Nice PR, thanks for that.

PR #667, opened 2021-04-10
Sync branches so v3 uses latest master

stephanebruckert:
@Peter-Schorn merging master into v3 was conflicting a bit, I think in the future it'll be easier to do the v2 change in v2 first, and then incorporate that change in v3 

Otherwise rebasing becomes difficult, as git isn't clever enough to detect that 2 different commits can have the same piece of code. Feature should be committed once, and then merged into other branches so they share the same references


Peter-Schorn:
Yes, I should've committed the non-breaking changes to master first. That was my mistake.

PR #679, opened 2021-04-22
Change assertGreater to assertGreaterEqual in some tests

gustavopcassol:
While running the tests, I got errors that were caused by the fact that my Spotify account didn't follow any artists, have any saved tracks nor have any top tracks. I changed `assertGreater` to `assertGreaterEqual` in the tests that were failing because of that, so that `0` is also valid.


gustavopcassol:
If this is not a good solution, I could change so that it saves a random track (for example) before running `assertGreater`, followed by unsaving it.

PR #691, opened 2021-06-16
Added DjangoSessionCacheHandler

varunpatil:
Added `DjangoSessionCacheHandler`, a cache handler that stores the token in the session framework provided by Django. Web apps using spotipy with Django can directly use this for cache handling.


stephanebruckert (approved):

stephanebruckert:
Nice, thank you!

Issue #439, opened 2020-02-12
Support for type hints


roeekl:
It would be nice to add a support for type hints. I know most of the functions have `str` parameters and returning `Dict` but it still can help in other functions like `prompt_for_user_token`.

stephanebruckert:
Hey @roeekl, looks like this is new in python3.5. If there is a way to add it while keeping compatibility with older python versions, we would definitely merge such PR!

IdmFoundInHim:
@stephanebruckert What is the minimum version of Python that spotipy 3 will need to support?

stephanebruckert:
@IdmFoundInHim I think we should choose what's easy for us. Developers should always think of upgrading their own version of python3. Starting at 3.7 sounds good to me

akathorn:
I use pylance strict type-checking in my project and the lack of type hints in Spotipy was irking me, so I started playing around with my own .pyi stubs using [stubgen](https://mypy.readthedocs.io/en/stable/stubgen.html) as a starting point.
I have annotated things lazily as I needed them, and soon I ran into the [problem of representing the type of the JSON](https://github.com/python/typing/issues/182) return values.
After playing around for a bit I realized that since Spotify responses have a fixed schema, they can be represented as [TypedDict](https://docs.python.org/3/library/typing.html#typing.TypedDict)s with really nice results. I scrapped Spotify's documentation and generated a TypedDict for each return type. Fun stuff.

I have a small example of the type-checking [here](https://github.com/akathorn/spotipy/blob/typings-example/test.py). 

My favorite thing about this approach is that vscode (and certainly other tools) autocompletes the fields of the response:
![Capture](https://user-images.githubusercontent.com/3438369/122816131-a18ae200-d2d6-11eb-98c4-966ba5d797ac.PNG)
Which means that you don't have to go to the Spotify docs or examine the dictionary at runtime to figure out what fields are inside, and what types they contain.
At runtime, the behavior is totally unmodified and they are still dicts. The new code is only used for type-checking.

Point is: if there is interest in this approach, I would be willing to go further with it and contribute it :)

Peter-Schorn:
@akathorn Open a pull request on the `v3` branch with these changes. Type hints are an incredibly useful feature.

akathorn:
I ran into an issue when representing Spotify JSON objects using `TypedDict`.
[...]

PR #695, opened 2021-06-23
Add type annotations

akathorn:
Working on #439. It is WIP but I would like some feedback before I validate the return types one by one.

All public methods are annotated. I haven't added anything that alters the runtime behavior in any way.
The arguments are straightforward since they are all basic types, but I had to define new types for the return values. The alternative is using `Dict[str, Any]` for all the JSON dicts, which would be more "lightweight" but provides near-zero typing information.

It works on nearly examples!


spotipy/json_types.py
```diff
@@ -0,0 +1,573 @@
...
+
+if sys.version_info >= (3, 8):
+    from typing import Literal, TypedDict
+else:
+    from typing_extensions import Literal, TypedDict
+
+_T = TypeVar("_T", covariant=True)
+
+
+TokenInfo = TypedDict("Token", {
```
Peter-Schorn, line 12: I would suggest changing `"Token"` to `"TokenInfo"` to match the name of the variable.

spotipy/json_types.py
```diff
@@ -0,0 +1,573 @@
...
+    "access_token": str,
+    "token_type": str,
+    "scope": str,
+    "expires_in": int,
+    "refresh_token": str,
+    "expires_at": int
+})
```
Peter-Schorn, line 19: `scope` and `refresh_token` will not be present for the client credentials flow, and so they should be optional.

spotipy/json_types.py
```diff
@@ -0,0 +1,573 @@
...
+    "is_local": bool,
+    "track": "Track",
+})
+
+PlaylistTracksRef = TypedDict("PlaylistTracksRef", {
+    "href": str,
+    "total": int,
+})
+
+PrivateUser = TypedDict("PrivateUser", {
```
Peter-Schorn, line 341: The `email`, `explicit_content`, `product`, and `country` properties of the `PrivateUser` object can still be `nil` if certain scopes are missing, and so these should also be optional. See [here](https://developer.spotify.com/documentation/web-api/reference/#object-privateuserobject).

spotipy/json_types.py
stephanebruckert: In the future we could look into creating a new type if possible, but probably not a worry now

akathorn: I discussed this [here](https://github.com/plamere/spotipy/issues/439#issuecomment-872464912). We can either:
- Ignore the fact that they will not be present
- Create a class `TokenInfo`
- Remove this, use `Dict[str, Any]`, and wait until `TypeDict` supports optional keys (or just forget it)

spotipy/json_types.py
akathorn: There is the same issue that I described [here](https://github.com/plamere/spotipy/issues/439#issuecomment-872464912).

PR #696, opened 2021-06-26
Update index.rst

hcientist:
Not sure how to edit RSt properly, but wanted to point out for others the superhelpful feature that spotipy already includes code to launch an http server to listen for the http callback


stephanebruckert (approved):
Thanks, I will check if the doc compiles when releasing a new version

Issue #652, opened 2021-03-04
Dumping improvement ideas for spotipy version 3


stephanebruckert:
Let's start thinking of the next major version for spotipy, while we keep maintaining the version 2.

spotipy's philosophy for v3 remains the same:

> A light weight Python library for the Spotify Web API

Light weight means we keep it simple and let developers parse the JSON results themselves, as we trust that the Spotify API is backward-compatible.

v3 PRs and issues can use the label:

[![](https://img.shields.io/github/labels/plamere/spotipy/spotipy3)](https://github.com/plamere/spotipy/labels/spotipy3)

## To do in version 3:

 - remove python2 support
 - support for type hints https://github.com/plamere/spotipy/issues/439
 - clean up deprecated methods
 - get rid of `oauth2.py` which confusingly includes non-oauth2 flows -> we could have an "auth" folder and a file for each authentication flow
 -  things tagged with [spotipy3](https://github.com/plamere/spotipy/labels/spotipy3)

Feel free to suggest anything that you believe could improve spotipy in the long term. Independent issues / PRs can still be created and referenced in this discussion.

## Development process

Here is a first suggestion, and can be improved:

 - a new [`v3` branch](https://github.com/plamere/spotipy/tree/v3) exists, so v3 PRs are welcome now! 
 - all backward-incompatible PRs can be merged into `v3`
 - backward-compatible PRs can be merged into `master` as usual
 - PRs welcome to regularly rebase `master` onto `v3` while making sure it doesn't include any of the things we don't want in v3
- we will be able to create pre-releases along the way https://www.python.org/dev/peps/pep-0440/#id27

----------

There is no rush, this will take the time it needs. Any suggestion is welcome! Thank you

Peter-Schorn:
I've noticed that the initializer for `Spotify` has three parameters that all alias each other: `client_credentials_manager`, `oauth_manager`, and `auth_manager`. We should remove the first two because they just create confusion.

Peter-Schorn:
Here's another idea: How about creating an Enum for all of the authorization scopes:

```python
from enum import Enum
from typing import Iterable
import re

class Scope(Enum):
    ugc_image_upload = "ugc-image-upload"
    user_read_recently_played = "user-read-recently-played"
    user_read_playback_state = "user-read-playback-state"
    playlist_modify_public = "playlist-modify-public"
    # and so on...

    @staticmethod
    def make_string(scopes: Iterable['Scope']) -> str:
        """
[...]

Issue #406, opened 2019-12-19
Choosing a new home for Spotipy


felix-hilden:
Hello again! Plamere has been inactive for a long time so Spotipy has been transferred to my name on [PyPI](https://github.com/pypa/pypi-support/issues/28) and [Read The Docs](https://github.com/readthedocs/readthedocs.org/issues/6335). In #404 we decided to continue the development of Spotipy. Now it's time to discuss where.

### Original issue: Is under development?
In August there was a long discussion in #387 on whether this project was abandoned or not. A number of alternatives were also declared as in development. I thought it would be fair to mention them here and include them in the list of considerations. In the order of announcement they were:

- [spotify-api](https://github.com/Saphyel/spotify-api) by Saphyel
- [spotipy](https://github.com/felix-hilden/spotipy) by felix-hilden
- [spotipy](https://github.com/Harrison97/spotipy) by Harrison97
- [spotipy](https://github.com/davidt99/spotipy) by davidt99

### The current discussion
We should discuss where the new home of Spotipy should be. That can be one of the repositories above or elsewhere if another one is announced. An appropriate time frame for the transfer should also be chosen, meaning a period where the information about the transfer is available but the package is still in its original state. The transfer is then announced in a separate issue.

State your case below. Please link to the repository, introduce it and argue why it should be chosen over other options that are presented.

**Please vote or comment on versions even if you have not written one!**

felix-hilden:
### [spotipy](https://github.com/felix-hilden/spotipy) - felix-hilden
I have written an updated version of the package. It's a complete client of the API in terms of authentication and endpoints, and has been in active development since August. We have an existing [update plan](https://github.com/felix-hilden/spotipy/issues/22#issuecomment-545008685) for the event that this repository is chosen. Not to mention the [cool logo](https://github.com/felix-hilden/spotipy/blob/master/docs/spotipy_logo_small.png) :D This is the kind of code you'll be writing:

```python
from spotipy import util, Spotify
from spotipy.scope import scopes
from spotipy.convert import to_uri

# Authenticate
cred = util.credentials_from_environment()
scope = scopes.user_library_read + scopes.user_modify_playback_state
token = util.prompt_for_user_token(*cred, scope=scope)

# Play a saved album
spotify = Spotify(token)
album = spotify.saved_albums(limit=1).items[0].album
album_uri = to_uri('album', album.id)
spotify.playback_start_context(album_uri)
```

#### Comparison with Plamere
[...]

Issue #404, opened 2019-12-10
I'm the new maintainer, what should we do?


felix-hilden:
Hello! Plamere has been inactive for a long time so Spotipy has been transferred to my name on [PyPI](https://github.com/pypa/pypi-support/issues/28) and a similar process is ongoing for [Read The Docs](https://github.com/readthedocs/readthedocs.org/issues/6335). Next, we should figure out what to do with the acquired rights.

### The situation
Plamere has been inactive since late 2017. The repository contains and the documentation references code that was not released in version 2.4.4, which is the most recent release on PyPI. This led to a lot of confusion for me when I tried out this library and I understand this has been the case for many others as well.

Spotify's Web API has progressed over the years. For example, playlists are no longer accessed via a user ([dev news](https://developer.spotify.com/community/news/2019/01/15/update-changes-to-playlist-uris/)) and new endpoints have been added. The API will continue to evolve.

Spotipy is [fairly popular](https://pypistats.org/packages/spotipy), pulling in up to a thousand downloads a day and gaining popularity over time. This means that lots of new users are affected by the decisions that are made now.

### The options
Possible solutions can be divided into two archetypes: conservative and progressive. I'll post them both below so we can discuss them in detail and tweak them if needed.

A number of forked/rewritten projects [have emerged](https://github.com/plamere/spotipy/issues/387). I have also written a version, which was what I referenced during the name transfers, but I'd like to keep the conversation open to other options as well. And I specifically would like to avoid choosing one of the new versions in this issue. If we choose the progressive solution, I'll open another issue for that discussion. That being said, please do dive into the issue and have a look at the proposed alternatives to gain an understanding of what types of solutions could be implemented as part of the progressive plan.

felix-hilden:
### Conservative
The conservative solution is the easy, hassle-free way out. In my mind it has three variations.

#### We do absolutely nothing
'Nuff said.

#### We publish one more version
Publishing the final contents of this repository on PyPI would bring the documentation in sync with the published version. This version could be `2.4.5` for example.

#### We notify users about other options
If maintenance is not continued, it could be fair to ease the pain of finding another up-to-date solution. This could be implemented in documentation, warning messages on importing the library or any other method.

In this case, we should publish the repo contents first in my opinion. Then if code needs to be modified, it could be version `2.5.0` for example.

Pros:
- Status quo is preserved. Nothing is required of existing users of the library.
- No confusion with new versions or other resources on the web being out-of-date.

Cons:
[...]

Issue #387, opened 2019-07-02
Is under development?


Saphyel:
Hello! I was thinking to check out this project but I don't see any activity in the last 2 years... is abandoned ?

idofr:
Had the same worries but I'm just starting to use it and it looks nice enough to use. Worst case you can always submit a merge request.

I already found what I think to be a small problem, but I'll start another discussion for that

marioortizmanero:
We should move everything to a fork since it really seems to be inactive. #211 is killing me. What fork is the most popular one?

Saphyel:
I was thinking to create a new project base on the new Spotify API in beta and python 3.7 ready.

pastaboy2131:
I just forked the project and a bit of help we could update it to a brand new one.

Saphyel:
There is like 400 forks already... Anyway I think will be best a new one that is easy to switch from this one as a 1st version. So I'll follow this approach.

felix-hilden:
It's nice to hear others having a similar concern. The author's email is not valid anymore. A bit ago I used this source to begin a small related project with the intention to include the library as a submodule, but seeing this discussion I think it's appropriate to pitch in my two cents. And I see @Saphyel and @pastaboy2131 have already started their own contribution!

There are lot's of forks, but if I'm not mistaken a large number of them have been used for pull requests for the original repo and such. At least I didn't find a fully up-to-date version.

A number of things bother me in the current state:
- The main class is rather long and hard to read
- There are loads of little inconsistencies and documentation errors
- Python 2 support seems useless now that it's retiring

And I'm sure you have your own. But I've tried to address these concerns in my reimplementation. It is still in its infancy, but next I plan to add types representing models returned by the Web API and proper documentation in the near future. And Saphyel's access token as a dataclass seemed like a great idea.

I would love some input on my code too, seeing there are other committed people here as well. Maybe we can get a longer discussion going.

Cheers!
- Felix

pastaboy2131:
@felix-hilden 
Thank you,
I would love to help and contribute to your project as well.

I'll be updating mine soon as well and I agree, the code is too long and hard to read, the author could have just created sub-modules for the main module.
What i am planning to do is to fully update the modules and make them more easier to read and/or modify.

felix-hilden:
That's terrific to hear!

[...]

PR #722, opened 2021-08-31
Objects instead of json

MakeShiftArtist:
# New Objects
Still a work in progress but I didn't like being forced to look at json data to see what was happening, so I added some objects to do that work for us. I did some bug testing and everything seems to be running smoothly. The objects should be fairly well documented as well
## User
Meant for user objects, not artists
```py
client.me()
```
Now returns a User object instead of some json data
## Artist
The artist object. Given that some info varies depending on how the request is made, i bug tested and it SHOULD work fine, but you may have some `None` values
## Album
The album object. there may be some None values, but there are 4 new methods to get additional info as well.
## Track
the track object, again there may be some None values but it has 3 new methods to get more infomation.

# New search methods
if a dev wants to search for artists, tracks, or albums, they can now use `client.search_artists`, `client.search_track`, and `client.search_albums`, respectively, which are generators that give the appropriate objects

# Example code
```py
import spotipy
import json
from spotipy.oauth2 import SpotifyOAuth

with open('config.json') as f:
  config = json.load(f)

client = spotipy.Spotify(
    auth_manager=SpotifyOAuth(
        client_id=config["client_id"],
        client_secret=config["client_secret"],
        redirect_uri="http://localhost/",
        scope="user-library-read"
    )
)

albums = client.search_albums("Perception", limit=1)
for album in albums:
    for track in album.get_tracks():
        print(track)
        print(track.explicit)
        break

"""
Intro III
False
"""
``

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,7 @@ extra_reqs = {
 
 setup(
     name='spotipy',
-    version='2.19.0',
+    version='2.19.1',
     description='A light weight Python library for the Spotify Web API',
     long_description=long_description,
     long_description_content_type="text/markdown",
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -13,6 +13,7 @@ import six
 import urllib3
 
 from spotipy.exceptions import SpotifyException
+from spotipy.objects import User, Artist, Album, Track
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: setup.py [hunk 1]
```diff
@@ -18,7 +18,7 @@
 
 setup(
     name='spotipy',
-    version='2.19.0',
+    version='2.19.1',
```
stephanebruckert, line 21: versions are bumped independently from the PRs

stephanebruckert (commented):
Hey! Sorry about the late review. I appreciate the work you put in in this PR but I see several issues with it:

 - structuring responses is nice, but to me it goes against the "light-weight" vision of spotipy where, we are supposedly confident to deal with raw responses in JSON. The benefit being that if the Spotify API publishes a new field, it will be available "as usual" and without having to modify the library. With the changes in this PR, I'm not sure we would benefit from that. I'm happy to be wrong about this and change my mind, but this is how I see it now
 - this looks backward-incompatible to me. If we are going to do this let's discuss it in the version 3 discussion https://github.com/plamere/spotipy/issues/652
 - the change only affects the search functions, while I believe if we were to do this, we should do it for all methods at once, everywhere

IdmFoundInHim:
The object-based structure is used by [Tekore](https://github.com/felix-hilden/tekore), another Spotify API wrapper under active development. I would recommend looking there, as this repo is committed to keeping the JSON output. (There was a [fairly](https://github.com/plamere/spotipy/issues/387) [substantial](https://github.com/plamere/spotipy/issues/404) [discussion](https://github.com/plamere/spotipy/issues/406) that led to the split of Tekore and Spotipy development to provide two different approaches.)

stephanebruckert:
Closing this as wont-fix.

Other than spotipy being light-weight and python being a straight-forward and to-the-point language, my main argument is that objectifying API responses can lead to making users believe a field isn't available when it is actually part of the returned response. Returning JSON will always tell "the truth".

PR #746, opened 2021-11-20
Adding example for unfollowing/deleting a playlist

rautskaa:
Add "delete_playlist.py" in examples.

### DIFF
--- a/examples/delete_playlist.py
+++ b/examples/delete_playlist.py
@@ -1,0 +1,25 @@
+import argparse
+import logging
+
+import spotipy
+from spotipy.oauth2 import SpotifyOAuth
+
+logger = logging.getLogger('examples.delete_playlist')
+logging.basicConfig(level='DEBUG')
+
+
+def get_args():
+    parser = argparse.ArgumentParser(description='Deletes a playlist for user')
+    parser.add_argument('-p', '--playlist', required=True,
+                        help='Playlist id')
+    return parser.parse_args()
+
+
+def main():
+    args = get_args()
+    sp = spotipy.Spotify(auth_manager=SpotifyOAuth())
+    sp.current_user_unfollow_playlist(args.playlist)
+
+
+if __name__ == '__main__':
+    main()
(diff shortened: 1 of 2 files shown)


examples/delete_playlist.py
```diff
@@ -0,0 +1,25 @@
...
+    parser = argparse.ArgumentParser(description='Deletes a playlist for user')
+    parser.add_argument('-p', '--playlist', required=True,
+                        help='Playlist id')
+    return parser.parse_args()
+
+
+def main():
+    args = get_args()
+    sp = spotipy.Spotify(auth_manager=SpotifyOAuth())
+    sp.current_user_unfollow_playlist(args.playlist)
```
stephanebruckert, line 21: this is `unfollow`, not`remove`

stephanebruckert (changes requested):

examples/delete_playlist.py
rautskaa: Based on the documentation https://developer.spotify.com/documentation/general/guides/working-with-playlists/, 
there is no endpoint for deleting a playlist and unfollow acts in the same way as removing a playlist:
"We have no endpoint for deleting a playlist in the Web API; the notion of deleting a playlist is not relevant within the Spotify’s playlist system. Even if you are the playlist’s owner and you choose to manually remove it from your own list of playlists, you are simply unfollowing it. Although this behavior may sound strange, it means that other users who are already following the playlist can keep enjoying it"

Should I update the file name to unfollow_playlist.py to make it more clear? Thanks.

stephanebruckert: I didn't know that but it makes sense because playlists are never deleted and can always be recovered.

> Should I update the file name to unfollow_playlist.py to make it more clear?

Yes please, if that's how Spotify defines it

rautskaa: Updated to unfollow_playlist.

stephanebruckert:
Thanks @rautskaa

PR #747, opened 2021-11-22
add RedisCacheHandler

ooyamatakehisa:
I used redis for token caching in my project, so I contribute this.
This may be useful when using multiple servers.
I'm not sure if I should include redis-py library to dependency of this project because redis library is not directly necessary for main functions of spotipy and only necessary for `RedisCacheHandler`.
May I have your help?

### DIFF
--- a/spotipy/cache_handler.py
+++ b/spotipy/cache_handler.py
@@ -1,10 +1,17 @@
-__all__ = ['CacheHandler', 'CacheFileHandler', 'DjangoSessionCacheHandler', 'MemoryCacheHandler']
+__all__ = [
+    'CacheHandler',
+    'CacheFileHandler',
+    'DjangoSessionCacheHandler',
+    'MemoryCacheHandler',
+    'RedisCacheHandler']
 
 import errno
 import json
 import logging
 import os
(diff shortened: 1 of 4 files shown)


stephanebruckert:
Thanks for that.

Looks like you will have to add the redis package in https://github.com/plamere/spotipy/blob/master/setup.py

Please also update the changelog

spotipy/cache_handler.py
```diff
@@ -138,3 +145,32 @@ def save_token_to_cache(self, token_info):
...
+        Parameters:
+            * redis: Redis object provided by redis-py library
+            (https://github.com/redis/redis-py)
+        """
+        self.redis = redis
+
+    def get_cached_token(self):
+        token_info = None
+        try:
+            token_info = json.loads(self.redis.get('token_info'))
```
Findus23, line 166: Won't `self.redis.get('token_info')` be `None` if the value hasn't yet been stored in Redis and then `json.loads(None)` throw an exception?

ooyamatakehisa: @Findus23 
You're right. I'm sorry for that. I'll fix it. 
Should I create a new branch and new pull request?

PR #751, opened 2021-12-05
Option to automatically return the complete playlist

### LINKED ISSUE #740: Option to automatically loop through paginated playlists with length more than 100 in spotipy.spotify.playlist() function
**Is your feature request related to a problem? Please describe.**
With the spotipy.spotify.playlist() function the items or tracks returned by default is limited to 100 and the results are paginated. And there is no way to read further details.

**Describe the solution you'd like**
Provide a parameter which would take care of looping through the paginated results and return the full playlist items.

**Describe alternatives you've considered**
An alternate solution that I ended up using is to get the total track number and loop over the results using spotipy.spotify,playlist_items with offsets at steps of 50 (max value allowed by spotify) to get the complete playlist items. Please see the code example below

```
sp = spotipy.Spotify(access_token)
results = sp.playlist(playlist_id)

total_track_count = results['tracks']['total']
max_limit = 50
offset = 0
remaining_track_count = total_track_count

tracks = []

while remaining_track_count > 0:
    results = sp.playlist_items(playlist_id,limit=max_limit,offset=offset)
    tracks.extend(results['items'])
    offset = offset + max_limit
    remaining_track_count = remaining_track_count - max_limit
```

RozayC:
This pull request resolve #740

Created a new function playlist_full in client.py that returns a complete list of tracks. Optional argument limit allows for less than the total track number within the playlist.
Add a test into integrations to check that the complete tracks are returned.


stephanebruckert:
I appreciate the effort (thanks!) but I don't think we should merge this. While it would be useful in most cases, it is a specific implementation which might not represent how everyone would like to retrieve a full list.

For example:
 - since a playlist can have up to 11k songs, one might want to add a logger to give feedback on the progress (fetched page 1/100, fetched page 2/100, etc.) 
 - someone who is building a website might want to show the first results before retrieving the whole list
 - someone who is worried about performance might want to start processing the first items before retrieving the whole list
 - someone who is searching for a song in a playlist, might want to stop retrieving the full list when a condition is met

This is likely the reason why the Spotify API itself doesn't have an endpoint to return the whole playlist.

Issue #757, opened 2021-12-16
Inconsistant result from playlist URI using Spotify.playlist_tracks


Darkfull-Dante:
**Describe the bug**
We use Spotipy for an Home Assistant custom component [Spotscat](https://github.com/fondberg/spotcast). An issue was raised by a user that seems to come from Spotipy handling of URI using the `?si=` element that was added by Spotify to comply with EU regulations. This means this bug is only reproduisible naturally from an EU based Spotify install. Otherwise, the `?si=` will not be present in the URI.

If a user pass a URI that was provided from an EU Spotify client, the playlist_tracks method will return a different dictionary than if it was provided without [see issue 267](https://github.com/fondberg/spotcast/issues/267):

| test              | result |
| ---------------: | --- | 
| with ?si=      |  dict_keys(['collaborative', 'description', 'external_urls', 'followers', 'href', 'id', 'images', 'name', 'owner', 'primary_color', 'public', 'sharing_info', 'snapshot_id', 'tracks', 'type', 'uri']) |
| without ?si= | dict_keys(['href', 'items', 'limit', 'next', 'offset', 'previous', 'total']) | 

**Your code**
```Python
import spotipy
from spotipy.oauth2 import SpotifyClientCredentials

auth_manager = SpotifyClientCredentials()
sp = spotipy.Spotify(auth_manager=auth_manager)

# uri with ?si=
result_1 = sp.playlist_tracks('spotify:playlist:5rAhD02CGReFtSbinTHxkn?si=968337997b194202')
print(result_1.keys())

# uri without ?si=
result_2 = sp.playlist_tracks('spotify:playlist:37i9dQZF1DZ06evO3hsjg5')
print(result_2.keys())

# potential fix
assert result_1['tracks'].keys() == result_2.keys()

# current uses in Spotcast
print(result_2['total'] - 1)
print(result_1['total'] - 1) # provides the max index of the tracks, this will fail
```

**Expected behavior**
Both types of URI should return the same behavior (a dictionary containing the tracks of a playlist)

**Output**
```Bash
>  python 267_test.py 
dict_keys(['collaborative', 'description', 'external_urls', 'followers', 'href', 'id', 'images', 'name', 'owner', 'primary_color', 'public', 'snapshot_id', 'tracks', 'type', 'uri'])
dict_keys(['href', 'items', 'limit', 'next', 'offset', 'previous', 'total'])
43
Traceback (most recent call last):
[...]

PR #758, opened 2021-12-16
Remove query parameters from URI

Darkfull-Dante:
#757

This PR adds a split to the _get_id function to make sure an URI provided by the user removes any element following an interrogation mark to only return the id element of the URI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 * Added `RedisCacheHandler`, a cache handler that stores the token info in Redis.
+* Changed URI handling in `client.Spotify._get_id()` to remove qureies if provided by error.
 
 ## [2.19.0] - 2021-08-12
 
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -12,8 +12,6 @@ import requests
 import six
 import urllib3
 
-import re
-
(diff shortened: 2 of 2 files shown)


Peter-Schorn:
The `si` query parameter is not part of the Spotify URI at all, so it definitely shouldn't be included in the result of the `_get_id` function.

Darkfull-Dante:
I saw that the current function does filter out queries but only from urls. The update I made to _get_id makes sure to also filter it out from URI.

With the `si`, Spotify is still able to process the URI and point to the correct playlist, I believe we should make sure to provide a fail proof experience to the potential user of the program, just like the user who reported an issue in the spotcast repo.

Peter-Schorn:
> I saw that the current function does filter out queries but only from urls.

That's because `si` is a query parameter that should only exist in URLs. How did you end up with `spotify:playlist:5rAhD02CGReFtSbinTHxkn?si=968337997b194202` in the first place?

Darkfull-Dante:
It seems that some of my users take the url, transform it themself in Uri, but forget to remove the queries from the original link.

I never had that problem personally, that's why we just got an issue last week for that and never had to bother with it before.

If I have some user that can do it, there might be other project with similar issue. Filtering out the query doesn't affect anything other than removing the bad element provided by the user.

stephanebruckert:
Yeah, this is more a usage issue as the URI is incorrect, but at the same time it wouldn't hurt to clean up the URI inside spotipy too.

Happy to merge this if you manage to fix the checks!

PR #760, opened 2021-12-20
added auth tokens placeholder

tonydavidx:
in the first 3 examples the SpotifyClientCredentials function does not show they require auth token as arguments so added a placeholder to show they need to input arguments before running the function.


stephanebruckert:
This is not needed because without it, it automatically looks for existing environment variables.

See https://github.com/plamere/spotipy/blob/d00b3d9ba2b1065352bdf598fd3028d7127944d6/docs/index.rst#client-credentials-flow

PR #761, opened 2021-12-22
Allow to set custom key in RedisCacheHandler

ENT8R:
I would like to use a different key than the one that is always set, so I thought it might be a good idea to allow users of the `RedisCacheHandler` to supply their own key and only set it to the default value of `token_info` if the key is not supplied as an argument.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 * Added `RedisCacheHandler`, a cache handler that stores the token info in Redis.
 * Changed URI handling in `client.Spotify._get_id()` to remove qureies if provided by error.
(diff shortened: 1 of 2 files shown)


stephanebruckert (approved):

stephanebruckert:
Looks good, please just add a changelog

ENT8R:
I added the changes to the changelog and I also added another check before trying to retrieve the value to make sure that the key exists (already noticed by @Findus23 in https://github.com/plamere/spotipy/pull/747#pullrequestreview-815915028)

stephanebruckert:
Thank you!

Findus23:
Sorry, to nitpick once again, but isn't something like 
```python
token_info = self.redis.get('token_info')
if token_info:
    return json.loads(token_info)
```
easier as it only does one request to redis instead of two (and doesn't have a really unlikely race-condition)?

stephanebruckert:
@Findus23 that's fair!

PR #765, opened 2022-01-03
Simplify check for existing token

ENT8R:
Follow-up of #761 which simplifies the check for an existing token in a Redis database based on https://github.com/plamere/spotipy/pull/761#issuecomment-999652928


stephanebruckert (approved):
Thanks!

PR #797, opened 2022-04-28
Fixed typo in code

tomowen02:
It said 'pprint' instead of 'print'


stephanebruckert:
That's not a typo, see https://github.com/plamere/spotipy/blob/50c4f8479513efe0545fd3da4b7ec9541522a962/examples/playlist_tracks.py#L3

Issue #753, opened 2021-12-09
albums endpoint should have country/market param


jadenmazzone:
Very simple but the albums(albums) endpoint should be albums(albums, country). According to Spotify Get Several Albums.

jadenmazzone:
Just "album" is also missing a market/country param.

stephanebruckert:
You probably refer to https://spotipy.readthedocs.io/en/2.19.0/#spotipy.client.Spotify.albums or https://spotipy.readthedocs.io/en/2.19.0/#spotipy.client.Spotify.album which looks for specific album(s) while the `market` parameter already exists on https://spotipy.readthedocs.io/en/2.19.0/#spotipy.client.Spotify.artist_albums, see `country`

jadenmazzone:
Those are different endpoints...Artist_albums does not give any track information while spotify albums / album does, therefore have different responses / data. https://spotipy.readthedocs.io/en/2.19.0/#spotipy.client.Spotify.albums and https://spotipy.readthedocs.io/en/2.19.0/#spotipy.client.Spotify.album should have a "country" param.

jadenmazzone:
Has this been added yet?

PR #800, opened 2022-05-01
Updated documentation about ISO-639 language codes and address issue #753

ivyadam:
Updated documentation about ISO-639 language codes to clarify that alpha-2 standard is used.


ivyadam:
Added `market` parameter to `album` and `albums` in spotipy.py to address ([#753](https://github.com/plamere/spotipy/issues/753).

stephanebruckert (approved):

PR #811, opened 2022-05-19
Fix text attribute for Response object

latenight01:
I got 
```
'Response' object has no attribute 'txt'
```
for a request and realized the correct attribute is [`text`](https://docs.python-requests.org/en/latest/api/#requests.Response.text) not `txt` for the `Response` object.


Peter-Schorn:
Good catch

stephanebruckert:
Thanks! Can you please add a changelog? In the meantime I will have a look at fixing the CI

mdmomeni:
I added a changelog yesterday

PR #824, opened 2022-06-08
Added Spanish support for selected pages

johnnybecerra:
- CONTRIBUTING, FAQ, and README files now can be viewed in Spanish

- modified orginal pages to include a link to view in Spanish

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,7 @@
 ## Contributing
 
+###### [Ver en español](https://github.com/plamere/spotipy/tree/master/espanol/FAQ_Espanol.md)
+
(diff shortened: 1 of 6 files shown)


johnnybecerra:
Spanish support to select pages

stephanebruckert:
I really appreciate the effort of the translation work and I understand some people might be more at ease reading Spanish.

However:
 - I believe this will be difficult to maintain in order to keep english/spanish docs synchronised. Will we have to review all translations when updating english?
 - There are also other main languages like russian or chinese that would deserve to be translated. If we start with spanish we will have to accept all languages. Do we want to spend time doing translation work rather than focus on the library?

Since programming is english, we should assume a minimum understanding of the english language. In the worse case, people can use Google Translate.

Issue #687, opened 2021-06-07
Error Only valid bearer authentication supported, reason: None??


phantom2152:
Hey I was getting info about a single track while I got this error

my code is

```
from spotipy import oauth2,spotify
re=oauth2.SpotifyClientCredentials(client_id=id,client_secret=sec).get_access_token()
spo = Spotify(re)
url = spo.track("https://open.spotify.com/track/3fHNjvF6AU1LXVez0wnFvw")
print(url)
```
And I am getting following error
```
HTTP Error for GET to https://api.spotify.com/v1/tracks/3fHNjvF6AU1LXVez0wnFvw returned 400 due to Only valid bearer authentication supported
Traceback (most recent call last):
  File "/home/anishgowda/.local/lib/python3.8/site-packages/spotipy/client.py", line 245, in _internal_call
    response.raise_for_status()
  File "/home/anishgowda/.local/lib/python3.8/site-packages/requests/models.py", line 943, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: https://api.spotify.com/v1/tracks/3fHNjvF6AU1LXVez0wnFvw

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/anishgowda/.local/lib/python3.8/site-packages/spotipy/client.py", line 339, in track
    return self._get("tracks/" + trid, market=market)
  File "/home/anishgowda/.local/lib/python3.8/site-packages/spotipy/client.py", line 291, in _get
    return self._internal_call("GET", url, payload, kwargs)
  File "/home/anishgowda/.local/lib/python3.8/site-packages/spotipy/client.py", line 261, in _internal_call
    raise SpotifyException(
spotipy.exceptions.SpotifyException: http status: 400, code:-1 - https://api.spotify.com/v1/tracks/3fHNjvF6AU1LXVez0wnFvw:
 Only valid bearer authentication supported, reason: None



```

Peter-Schorn:
Read the documentation: https://spotipy.readthedocs.io/en/2.18.0/#client-credentials-flow

Peter-Schorn:
> ```
> spo = Spotify(auth=re['access_token'])
> ```

While this works, it is not recommended because the access token will not be automatically refreshed when it expires after an hour. As shown in the README, this is how you should initialize `Spotify`:
```
[...]

Issue #828, opened 2022-06-11
Breaking change in minor version?


SHxKM:
Hi,

Super glad to see this maintained after a long halt.

I've noticed after upgrading from `2.4.4.` to `2.19.0` that some endpoints like `search` no longer work when the `Spotipy` object is initialized with a `auth` param. 

My old code (now returns an error):

```python
credentials = oauth2.SpotifyClientCredentials(
        client_id=SPOTIFY_CLIENT_ID,
        client_secret=SPOTIFY_SECRET_ID,
    )
token = credentials.get_access_token()
spotipy_obj = spotipy.Spotify(auth=token)
artist_on_spotify = spotipy_obj.search(q="X", type="artist", limit=1)
```

The new code:

```python
spotipy_obj = spotipy.Spotify(client_credentials_manager=credentials)
artist_on_spotify = spotipy_obj.search(q="X", type="artist", limit=1)
```

Meanwhile, user-related endpoints continue to work with `token`.

Is this expected? I couldn't find anything in the release notes regarding this. Maybe I didn't look hard enough.

Also: am I correct in saying that the cache system is only for the application's access token, not those of the application's users?

Peter-Schorn:
>My old code (now returns an error):
>
>```
>credentials = >oauth2.SpotifyClientCredentials(
>        >client_id=SPOTIFY_CLIENT_ID,
>        >client_secret=SPOTIFY_SECRET_ID,
    )
>token = >credentials.get_access_token()
>spotipy_obj = >spotipy.Spotify(auth=token)
>artist_on_spotify = >spotipy_obj.search(q="X", >type="artist", limit=1)
>```

This will always return an error after an hour when the access token expires. The refresh token has not been saved.

Therefore, this pattern is _never_ recommended.

SHxKM:
@Peter-Schorn that code is just a simplified example. I’m not complaining that it’s changed I just want to know if it indeed was.

Peter-Schorn:
Your code is flawed for any version as already described.

SHxKM:
> Your code is flawed for any version as already described.

You can repeat yourself being condescending all you want. The question remains: a code that works in 2.4.4 doesn’t work 2.19.0, is this breaking change intended?

Peter-Schorn:
Don't know.

SHxKM:
Hi @stephanebruckert, thank you. With `2.19.0` here's the exception of type `spotipy.client.SpotifyException`:

```
[...]

PR #832, opened 2022-06-19
Fixes "Only valid bearer authentication supported, reason: None"

stephanebruckert:
Fixes https://github.com/plamere/spotipy/issues/687, https://github.com/plamere/spotipy/issues/828

------------

To keep backward-incompatibility we shouldn't have made `as_dict=True` the default in `get_access_token` methods. This PR swaps it back to `False`. Therefore this must be seen as a bugfix rather than a new backward-incompatible change.

Normally, this shouldn't have too much effect on new code as we now don't recommend to get the token independently via `get_access_token()` in order to authenticate (see README).

If you have been using `get_access_token()` in new code, you should now use `get_access_token(as_dict=True)`.


stephanebruckert:
Actually I'm going to close this, I think I got it wrong. From this PR https://github.com/plamere/spotipy/pull/435/files it looks like we were always returning a dict.

PR #839, opened 2022-07-15
fixed raw_input bug

Gauds:
changed raw_input to input since raw_input doesn't work with python 3

### DIFF
--- a/spotipy/oauth2.py
+++ b/spotipy/oauth2.py
@@ -112,7 +112,7 @@ class SpotifyAuthBase(object):
     @staticmethod
     def _get_user_input(prompt):
         try:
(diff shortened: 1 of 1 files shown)


spotipy/oauth2.py
```diff
@@ -112,7 +112,7 @@ def redirect_uri(self, val):
     @staticmethod
     def _get_user_input(prompt):
         try:
-            return raw_input(prompt)
+            return input(prompt)
```
stephanebruckert, line 115: this is exactly what's in the `except`?

stephanebruckert:
Your change doesn't make much sense, I think you got confused.

PR #840, opened 2022-07-29
Fixed id recognize

albertopasqualetto:
Probably api is changed

### DIFF
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -1917,7 +1917,8 @@ class Spotify(object):
         return path
 
     def _get_id(self, type, id):
-        id = id['uri']
+        if (isinstance(id,dict)) and ('uri' in id) and (id['uri']):
+            id = id['uri']
         fields = id.split(":")
         if len(fields) >= 3:
             if type != fields[-2]:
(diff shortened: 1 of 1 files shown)


stephanebruckert:
Please provide more details, I don't understand your change and I am not aware of any API changes. Please also try to run lint and unit tests locally as this is currently failing. Check https://github.com/plamere/spotipy/blob/master/CONTRIBUTING.md

albertopasqualetto:
Well, I fixed flake8 errors.
Anyway I get more failures (9) with no changes than with my change (7); the error I get is this:
Immagine 2022-08-02 160323
Because `playlist_id` is a dictionary in this case, so the actual id needs to be extracted from the dict.

albertopasqualetto:
I see that you are deprecating the 'playlist_add_items` using IDs with #919 , so maybe this PR could be closed, since this error appears in this case.

But if nothing has changed, this was not working since my PR (try testing the playlist_add_items(id)) and this would be until v3 #914 .

[This is my code](https://github.com/albertopasqualetto/spotify-playlists-mover/blob/41a8bc29f0d84ea2982978230e5fe5d54cabace2/spotify_playlists_mover/moving_methods.py#L79)

stephanebruckert:
Ok I'm going to close it.

In your code instead of:

    sp_to.playlist_add_items(new_playlist['id'], chunk)

please just use:

    sp_to.playlist_add_items(new_playlist['uri'], chunk)

PR #841, opened 2022-08-02
Fix linting in test_non_user_endpoints which caused failed builds in flake8 step

albertopasqualetto:


browniebroke (commented):
I just tried to contribute and I had the same issue. Looks good to me, but I'm wondering (maybe one for the maintainer) why we needed the parentheses in the first place? They seem redundant to me...

PR #849, opened 2022-09-02
Added `Spotify.artist_image` and `Spotify.album_cover_image` functions to `client.py`

erie-lake-erie:
Note: `Spotify.artist_image` and `Spotify.album_cover_image` return the image as a list instead of a dictionary to stay in line with `Spotify.playlist_cover_image`, which returns the image as
```python
[
  {
    'height': either_integer_or_None,
    'url': 'https://i.scdn.co/image/.....',
    'width': either_integer_or_None
  }
]
```

### DIFF
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -397,6 +397,29 @@ class Spotify(object):
             offset=offset,
         )
 
+    def artist_image(self, artist_id, idx=0, fallback=True):
+        """Retrieves the artist image.
+
+           Parameters:
+               - artist_id - the artist ID, URI, or URL
+               - idx - index of the image array. Smaller index means larger image.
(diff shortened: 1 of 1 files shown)


stephanebruckert:
Appreciate the effort but I don't think we want to create a method for each specific field. If we were to merge it that would mean we could also create `artist_popularity()` which would be endless :)

For a minimalistic and to-the-point library like spotipy it should be fine for everyone to just use `artist()['images']`

Peter-Schorn:
Also note that the reason `Spotify.playlist_cover_image` exists is that there's a specific [endpoint](https://developer.spotify.com/documentation/web-api/reference/#/operations/get-playlist-cover) for it.

PR #855, opened 2022-09-16
Example Typo "pprint"

FullByte:


stephanebruckert:
it's not a typo, see [`pprint`](https://docs.python.org/3/library/pprint.html) package:

https://github.com/plamere/spotipy/blob/0d8fb4896d055cf572e312d03509f2c6ff860f0a/examples/playlist_tracks.py#L3

PR #869, opened 2022-10-12
Make redis dependency optional

ahesford:
The redis package is only used when somebody uses the redis cache handler, so the imports can be restructured to make redis an optional dependency.


ahesford:
Ping

Peter-Schorn:
We could catch the import error and display a detailed error message indicating that redis is an optional dependency and must be explicitly installed (clients often expect pip install -r requirements.txt to install all the dependencies for a package).

ahesford:
For starters, I added a note to the readme in the latest commit.

stephanebruckert:
> We could catch the import error and display a detailed error message indicating that redis is an optional dependency and must be explicitly installed (clients often expect pip install -r requirements.txt to install all the dependencies for a package).

True, exceptionally for this use case of "connectors" we should go with the warning. @ahesford this is how a warning can be raised https://github.com/plamere/spotipy/blob/fa44fed76d7aecb70f9ee6cde14da5e2ce1e17d7/spotipy/client.py#L1951

> For starters, I added a note to the readme in the latest commit.

Thanks that helps. Note that in any case we won't be able to merge this into v2 aka master just now, because it would break existing apps. So this will have to go into v3 https://github.com/plamere/spotipy/tree/v3
See https://github.com/plamere/spotipy/issues/652
Can you please change the base of your PR from `[plamere:master](https://github.com/plamere/spotipy)` to `[plamere:v3](https://github.com/plamere/spotipy/tree/v3)`? Feel free to create a new PR if it gets messy

ahesford:
Rebased on v3. That branch had no redis cache handler at all, so I brought the whole class over.

I added an import check during creation of a cache object to make sure redis is available and warn if not. I'm not sure there would be a good way to handle this in the functions where the import are actually needed, unless the exception catching is moved from `RedisError` to `Exception` to issue a blanket warning instead of hard errors for unexpected, non-redis failures.

spotipy/cache_handler.py
```diff
@@ -107,3 +108,45 @@ def get_cached_token(self):
...
+            * key: May be supplied, will otherwise be generated
+                   (takes precedence over `token_info`)
+        """
+        self.redis = redis
+        self.key = key if key else 'token_info'
+
+        try:
+            from redis import RedisError
+        except ImportError:
+            warnings.warn('Error importing requied module "redis"', UserWarning)
```
Peter-Schorn, line 132: There's a typo here: "requied".

Peter-Schorn: Also, the word "required" may imply to some users that this module is required for the whole package. Instead, the error message should indicate that this module is required *for the `RedisCacheHandler`*.

ahesford: Fixed. I also restructured the history to preserve the evolution of Redis support on the master branch and diminish the extent of my own changes.

spotipy/cache_handler.py
```diff
@@ -107,3 +113,49 @@ def get_cached_token(self):
...
+        """
+        self.redis = redis
+        self.key = key if key else 'token_info'
+
+        try:
+            from redis import RedisError
+        except ImportError:
+            warnings.warn(
+                'Error importing module "redis", '
+                'required for RedisCacheHandler',
```
Peter-Schorn, line 139: This is a comma splice. I suggest changing it to `Error importing module "redis"; it is required for RedisCacheHandler`

ahesford: Done

stephanebruckert (approved):

PR #889, opened 2022-11-16
Provide response type and example output to code examples

TroyPeele2:


stephanebruckert:
Answered in https://github.com/spotipy-dev/spotipy/pull/882

PR #701, opened 2021-07-01
Add type annotations

akathorn:
All parameters for public methods are annotated.
Methods that return a JSON are annotated with a return type of `Dict[str, Any]`, which is quite limited.

This is part of #439.

PR #895, opened 2022-11-26
Add parameter type hinting to API reference

kevinsekuj:


stephanebruckert:
Indeed type hinting is something we will need at some point, however since spotipy 2.x still supports python 2.7, I wonder if type hinting isn't going to break on 2.7? If it is breaking it that's fine, we can still add type hinting in spotipy 3.x, see https://github.com/spotipy-dev/spotipy/issues/652. In that case the PR must be raised against the v3 branch. However please be aware there is already https://github.com/spotipy-dev/spotipy/pull/701 and https://github.com/spotipy-dev/spotipy/pull/695

kevinsekuj:
I see, thanks for the info. I was aiming more towards adding typing to the API reference documentation, using this package for sphinx, which auto generates the typing in the docs based on the source code. 

So when adding type hints to function parameters and setting up the docstrings correctly, the docs build looks something like

image

I see that #701 uses more thorough types from the `typing` package, fortunately sphinx-autodoc-typehints should be compatible with those too.

PR #899, opened 2022-11-28
update TUTORIAL.md

alhamilton1111:
Recently went through the setup process and noticed some missing/updated information that would be useful for beginners. Check it out and let me know what you think. Thanks!


TUTORIAL.md
```diff
@@ -22,15 +22,19 @@ This tutorial will be easiest if you have some knowledge of how to use Linux com
 Once those three setup items are taken care of, you're ready to start learning how to use Spotipy!
 
 ## Step 1. Creating a Spotify Account
-Spotipy relies on the Spotify API. In order to use the Spotify API, you'll need to create a Spotify developer account.
+Spotipy relies on the Spotify API. In order to use the Spotify API, you'll 
+need to create a Spotify developer account. Here is a how-to [video](https://www.youtube.com/watch?v=3RGm4jALukM) that showcases how to get
```
stephanebruckert, line 26: this video might be useful but the image is very low quality and hard to read. I would rather not include this in official docs.

TUTORIAL.md
```diff
@@ -22,15 +22,19 @@ This tutorial will be easiest if you have some knowledge of how to use Linux com
...
+A. Visit the [Spotify developer portal](https://developer.spotify.
+com/dashboard/). If you already have a Spotify account, click "Log in" and enter your username and password. Otherwise, click "Sign up" and follow the steps to create an account. After you've signed in or signed up, you should be redirected to your developer dashboard. You do not need to have a premium Spotify account, a free account will work.
 
 B. Click the "Create an App" button. Enter any name and description you'd like for your new app. Accept the terms of service and click "Create."
 
-C. In your new app's Overview screen, click the "Edit Settings" button and scroll down to "Redirect URIs." Add "http://localhost:1234" (or any other port number of your choosing). Hit the "Save" button at the bottom of the Settings panel to return to you App Overview screen.
+C. In your new app's Overview screen, click the "Edit Settings" button and scroll down to "Redirect URIs." Add "[http://localhost:1234/callback/]" (or any other port number of your choosing). Hit the "Save" button at the bottom of the Settings panel to return to you App Overview screen.
```
stephanebruckert, line 34: The callback URL can be anything. While having `/callback` in it would work, it would also work without it.

stephanebruckert:
Appreciate the improvement although I'm going to close it because it relies on that video which is only a possible implementation. Also the video is LQ.

PR #900, opened 2022-11-28
Fix typo in "pprint(res)"

willtholke:


stephanebruckert:
it's not a typo, see
https://github.com/spotipy-dev/spotipy/blob/ea3507bebc233ccd0546d635b30386ed3506a3f2/examples/player.py#L3

PR #910, opened 2022-12-01
Documentation update for README.md, bug_report.md, and TUTORIAL.md

lucasj9:
Added a bug reporting section to the README.md so contributors can collect the information before submitting a bug report.

This should also help filter out some of the bug reporting as the descriptions for each of the sections is more detailed and hopefully provokes contributors to sig in a bit more and provide more information

Also to address an issue I saw posted, I referred to the setup.py to add a line in the tutorial.md that specifies that any python version above 3.4 will work with the library. This is a small issue to fix but for some that are just starting with python the version numbers can be confusing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,4 +68,42 @@ Don’t forget to add the *Spotipy* tag, and any other relevant tags as well, be
 
 If you have suggestions, bugs or other issues specific to this library,
 file them [here](https://github.com/plamere/spotipy/issues).
-Or just send a pull request.
+Or just send a pull request. 
+
+Refer to Documentation below for collecting information for bug reporting.
+
+## Bud Reporting
+
+### Describing the Bug 
+
+Try to explain in detail what the bug is and how it is caused.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -58,14 +58,54 @@ for idx, item in enumerate(results['items']):
...
 
-If you have suggestions, bugs or other issues specific to this library,
+If you have suggestions or other issues specific to this library,
 file them [here](https://github.com/plamere/spotipy/issues).
-Or just send a pull request.
+Or just send a pull request. 
+
+Refer to Documentation below for collecting information for bug reporting.
+
+## Bud Reporting
```
stephanebruckert, line 75: This doesn't need to be in the README. Please create a PR to update this file directly if you think anything is missing in there .github/ISSUE_TEMPLATE/bug_report.md

stephanebruckert (changes requested):

PR #914, opened 2022-12-05
updated functionality of playlist_add_items()

oliveraw:
playlist_add_items() should not be able to accept pure IDs, since the type 'track' or 'episode' cannot be inferred. We are proposing a change where playlist_add_items is modified to only accept URIs and URLs


oliveraw:
The issue was first mentioned in https://github.com/spotipy-dev/spotipy/issues/810, wherein episodes could not be added to playlists by ID or URL (only by URI). The proposed solution allows for both track and episode URIs and URLs, but no IDs. We have also added another (deprecated) function, user_playlist_add_episodes(), in order to match the existing user_playlist_add_tracks().

stephanebruckert:
Thanks for that!

That's a fairly annoying bug as the fix would be backward-incompatible. If merged now it will likely break things for a majority of apps dealing with plain track ids and fix things for only a minority of apps dealing with plain episode ids. That's probably a good reason to keep it for the next major version ([v3](https://github.com/spotipy-dev/spotipy/issues/652)) meaning it can't be merged in the current master branch (v2).

To get the best of both world my suggestion would be to open 2 PRs:

- one against the current master branch that we will then release as v2.23.0 :
  - leave `playlist_add_items`'s functionality as it is but update the docstring to let people know that episodes IDs are expected to be full URIs
   - still add the new deprecated `user_playlist_add_episodes` so that people can use the feature without full URIs
- one against the [`v3` branch](https://github.com/spotipy-dev/spotipy/tree/v3) that we will release as v3.0.0:
  - fix `playlist_add_items` to handle both as done in the current PR
  - no need for the deprecated ones anymore

The first PR would be optional because only informative, but if doing both my recommendation is to start with the first one, and then base the v3 update from it.

oliveraw:
sounds good, will keep an eye out!

stephanebruckert:
@oliveraw I just updated v3 https://github.com/spotipy-dev/spotipy/tree/v3

stephanebruckert (approved):

PR #916, opened 2022-12-07
Add open_browser kwarg for prompt_for_user_token

### LINKED ISSUE #792: Using util.prompt_for_user_token on a headless machine, can't use open_browser=False?
Okay, so I've recently moved on to a Raspberry Pi although whenever I try to use Spotipy it gets stuck.
I figured out that the problem is that normally, a browser window would open so I can authenticate my program.
I'm now working on a headless machine so it of course doesn't work.
There has been an issue like this before and it was said that there was the variable open_browser added, which is exactly what I'm looking for.
Although with my method of authorisation, I can't use open browser?

The only pieces of code I have are

`token = util.prompt_for_user_token(username=username,scope=scope, client_id=client_id_spotify, client_secret=client_secret_spotify, redirect_uri=redirect_uri)`

and then

`sp = spotipy.Spotify(auth=token)`

The problem is that open_browser can't be used with util, and I have no idea how to get it working.

joshhoughton:
When attempting to authenticate via `spotipy.util.prompt_for_user_token()` in a headless environment (where `webbrowser` fails), it's not currently possible to authenticate.

These small changes allow `open_browser=False` to be passed to instead print out the authentication URL so that it can be accessed in another environment where a browser is accessible.

First PR here, feedback appreciated!

Fixes #792

### DIFF
--- a/spotipy/util.py
+++ b/spotipy/util.py
@@ -28,7 +28,8 @@ def prompt_for_user_token(
     redirect_uri=None,
     cache_path=None,
     oauth_manager=None,
-    show_dialog=False
+    show_dialog=False,
+    open_browser=True
 ):
     warnings.warn(
         "'prompt_for_user_token' is deprecated."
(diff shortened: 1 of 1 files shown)


stephanebruckert:
While we could merge this, `spotipy.util.prompt_for_user_token()` is deprecated and it is now recommended to use:

    spotipy.Spotify(auth_manager=SpotifyOAuth(open_browser=False))

See https://github.com/spotipy-dev/spotipy/blob/master/examples/headless.py for an example

joshhoughton:
Damn, missed that! I'll give it a try!

PR #964, opened 2023-04-12
Ignore Attribute error in Spotify.__del__

johncheng2011:
I was trying to create a class that Inherits spotipy.Spotify. But found this interaction where if an exception is raised in the child class's `__init__` it will try to run `Spotify.__del__` where `self._session` hasn't been defined yet. So it raises an Attribute Error instead of the Exception from the child class.
 

For example
```
from spotipy import Spotify


class Spotify2(Spotify):
    def __init__(self):
        raise Exception
    

spotify_instance = Spotify2()
```
Will result in this error ignoring the Exception raised in the child class's `__init__`
```
  File "/Users/Me/Desktop/test.py", line 22, in <module>
    spotify_instance = Spotify2()
  File "/Users/Me/Desktop/test.py", line 19, in __init__
    raise Exception
Exception
Exception ignored in: <function Spotify.__del__ at 0x101b7df70>
  File "/Users/Me/.pyenv/versions/3.9.15/lib/python3.9/site-packages/spotipy/client.py", line 214, in __del__
    if isinstance(self._session, requests.Session):
AttributeError: 'Spotify2' object has no attribute '_session'
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Changes the YouTube video link for authentication tutorial (the old video was in low definition, the new one is in high definition)
 - Updated links to Spotify in documentation 
-- Fixed error obfuscation when Spotify class is being inherited and an error is raised in the Child's __init__
+- Fixed error obfuscation when Spotify class is being inherited and an error is raised in the Child's `__init__`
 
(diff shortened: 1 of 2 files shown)


dieser-niko (approved):
Love it.

dieser-niko:
Wait, before I merge, what is this about?
https://github.com/spotipy-dev/spotipy/blob/6d8d9d1f9ba578ff3e515adb305d0dabb9c4eb07/CHANGELOG.md?plain=1#L22

PR #976, opened 2023-05-26
documentation work

chrishughes20:
I have edited some of the documentation to hopefully make a few things more clear for users of Spotipy as well as future contributors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Changed
+- Edited TUTORIAL.md, CONTRIBUTING.md, and TUTORIAL.md, fixed/added several links, and added clarification for some steps
+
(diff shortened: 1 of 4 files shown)


dieser-niko (changes requested):
Many changes seem unnecessary or excessive.
Just to list some examples:

`CONTRIBUTING.md`:
- explanation on what environment variables are (I'd expect a contributor to know about this already)
- verifying if variables have been set correctly (same as above)
- console commands have been moved a bit to the right (why?)

`README.md`:
- converting a link to a hyperlink whose text is exactly the same as the link (markdown does this for you, no need for `[text](link)`)
- "Or upgrade PIP if needed" whose command does not actually upgrade PIP

However, there are some changes that I do like. `TUTORIAL.md` has some nice additions like highlighting variables, correcting spelling mistakes and maintaining consistency. There's just one thing I'd like to point out: For example, when you highlight variables, you are using \```variable```. Please use \`variable\` instead (except for code blocks where it's necessary)

PR #985, opened 2023-06-07
Updated ReadTheDocs page to make site more beginner friendly

CodeZ89:
I have updated the documentation in the ReadTheDocs site to make the site a little more user friendly. The tutorial that is on the repository I could not find linked anywhere, so I added a link to the page as well as put some of that information directly in the docs. I also added information on setting up an app on the Spotify Web API, so a first time user does not have to navigate off the page to get the general idea of how to do so. I reorganized some of the content in the docs to better flow with the additional information. I would really appreciate any honest feedback about my changes and how I may improve in the future. 

In addition, I updated the CHANGELOG documentation to make it more user friendly as well. For those users who have never contributed or used a changelog before, it gives them a better idea of why and how it should be used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 Changelog Basics & Principles
 -----------------------------
+Before adding to the Changelog, think about the following: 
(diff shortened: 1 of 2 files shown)


stephanebruckert:
changelog and python basics shouldn't be part of this repo

PR #992, opened 2023-06-12
Update video tutorial

paulreyneveld:


stephanebruckert:
In the future we will focus on better written doc rather than redirecting to videos

PR #1013, opened 2023-08-15
Created EXAMPLES.md and audiobook support

kkowalsks:
Created an organized EXAMPLES.md file for new users to have easier access to functionality.
Added audiobook ID, URI and URL support and an example function `show_audiobook.py` for usage.


stephanebruckert:
Thanks but will close this:

 - audiobook support was added in another PR
 - we will now use this repo only for examples as they are better organised there https://github.com/spotipy-dev/spotipy-examples/blob/main/showcases.ipynb

PR #1015, opened 2023-08-16
Updated contributing to include documentation guidelines

nealchapman1995:


stephanebruckert:
That shouldn't be part of the spotipy docs

PR #1042, opened 2023-11-22
Added MemcacheCacheHandler

andrewcara:
Added a CacheHandler for Memcached using pymemcache.


dieser-niko:
I need your thoughts on this because I've never heard of memcache. Is it popular? And does it make sense to use with spotipy?
This needs to be considered because it would mean that we would have to add another requirement to our list.

stephanebruckert (commented):
Memcache is widely used and just like the other cache handlers would allow to store the token somewhere else.

The new requirement is a problem, so perhaps we should move it to `extras_require ` and update the installation instructions to look like https://stackoverflow.com/a/65936306/1515819

However we haven't done that for Redis. We could leave Redis where it is for now and move it to extras in v3 to maintain backward compatibility.

stephanebruckert (approved):

Issue #936, opened 2023-01-28
Hidden limits for current_user_saved_* endpoints 


rv3392:
**Describe the bug**
When doing a request to some (potentially all, but I haven't tested) of the current_user_saved_* endpoints such as current_user_saved_shows_add and current_user_saved_shows_delete there is a hidden limit (of 50) to the number of IDs that can be passed in. This is a restriction on Spotify's end, but isn't really documented anywhere in the library. I think simply documenting this should be enough to avoid any errors, it should just be clear that passing more than 50 items to these endpoints will cause an error.

**Your code**
Code that doesn't work (passing 51 items to the current_user_saved_shows_add function):
```
import spotipy
from spotipy import SpotifyOAuth

[...]

PR #1050, opened 2023-12-07
Addressed issue #936

lishukie6588:
Added an additional FAQ addressing the hidden limit of 50 shows per request for the API endpoints related to accessing/adding/deleting current user saved shows suffixed by "current_user_saved_shows" in client.py, resulting in an unexplained HTTP error response.

https://github.com/spotipy-dev/spotipy/issues/936

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Changes the YouTube video link for authentication tutorial (the old video was in low definition, the new one is in high definition)
 - Updated links to Spotify in documentation 
 
+## [2.23.1] - 2023-10-30
+
+### Added
+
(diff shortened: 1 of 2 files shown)


dieser-niko:
I'm not sure about this.

This limit is not hidden. It is clearly documented in Spotify's Web API for each parameter and is being used as a default value in the spotipy library.

If a programmer missed these mentions, they're likely to miss the FAQ.

dieser-niko:
I just noticed that stephanebruckert has changed the tags on the mentioned issue. I'll keep this open until this is resolved.

stephanebruckert:
Ok let's close this. After reviewing, it seems the issue in [#936](https://github.com/spotipy-dev/spotipy/issues/936) can be addressed with updates to docstrings, code, or warnings. No need to add it to the FAQ; the issue itself is sufficient

PR #1065, opened 2023-12-28
Drop support for EOL Python 3.7

hugovk:
Remove redundant dependencies and code, and upgrade syntax using pyupgrade.


dieser-niko:
Looks good, but just one question: You have replaced `str.format()` with formatted strings (`f""`), but in `oauth2.py` you're using `str.format()` again. Is there a reason for that?

hugovk:
I used pyupgrade which is "intentionally timid": https://github.com/asottile/pyupgrade#f-strings

I've upgraded some more to use f-strings, using https://github.com/ikamensh/flynt

dieser-niko:
@stephanebruckert I've gone through the code, but I'd like your OK as there are a lot of changes.

stephanebruckert (approved):
Looks good!

Issue #1066, opened 2023-12-31
Example code does not work


aryan-gupta:
**BUG**
A script I wrote few months ago is not working. To debug this issue, I ran the example: `examples/user_playlists_contents.py` which also failes with an error similar to my code.

**MY CODE**
This code gets the list of tracks from all the playlists the user has. `user_playlists_short` is a list of two-tuples with first element being the name of the playlist and the second item being the playlist id. This fails with this error:
```
def get_tracks(user_playlists_short):

    user_tracks = []
    user_tracks_set = set()
    for name, plid in user_playlists_short:
        results = sp.playlist(plid, fields="tracks,next")
        tracks = results['tracks']
        tracks_json = tracks['items']


        while tracks['next']:
            tracks = sp.next(tracks)
            tracks_json.extend(tracks['items'])
        
        user_tracks = user_tracks + [ (i['track']['name'], i['track']['id']) for i in tracks_json ]
        user_tracks_set = user_tracks_set.union(set([ i['track']['id'] for i in tracks_json ]))

    return user_tracks_set
```
ERROR:
```
Traceback (most recent call last):
  File "/[omitted]/spotify/./match_friends.py", line 94, in <module>
    ari_tracks = get_tracks(ari_playlists_short)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/[omitted]/spotify/./match_friends.py", line 77, in get_tracks
    tracks_json.extend(tracks['items'])
                       ~~~~~~^^^^^^^^^
KeyError: 'items'
```

The correct output should not error and the `tracks` inside the while loop should contain the next paged results. However, probing the tracks variable, there is only one item, but should contain the next paged link and the current list of tracks. 

```
{'next': 'https://api.spotify.com/v1/playlists/[playlist_id_omitted]/tracks?offset=200&limit=100&fields=tracks,next&additional_types=track'}
```

**FULL MINIMAL EXAMPLE**
This is the `examples/user_playlists_contents.py` example with some modifications for the environment variables. This also errors out similar to my code. 
```
import string
import time
import os
from os.path import join, dirname
from dotenv import load_dotenv
import json
import spotipy
from spotipy.oauth2 import SpotifyClientCredentials
from spotipy.oauth2 import SpotifyOAuth

dotenv_path = join(dirname(__file__), 'secret.env')
load_dotenv(dotenv_path)


[...]

PR #1068, opened 2024-01-02
Fix broken example

Jack-Dane:


Jack-Dane:
Fixes an issue with the example raised in #1066.

dieser-niko:
I would go a different way. The reason this fails at all is because the next url is `/playlists/<playlist_id>/tracks` while the initial request url is `/playlists/<playlist_id>`. It should be more straight forward with `sp.playlist_tracks` instead of `sp.playlist` so we can get a consistent response.

examples/user_playlists_contents.py
```diff
@@ -25,7 +25,7 @@ def show_tracks(results):
             print(playlist['name'])
             print('  total tracks', playlist['tracks']['total'])
 
-            results = sp.playlist(playlist['id'], fields="tracks,next")
+            results = sp.playlist(playlist['id'], fields="tracks,next,items")
             tracks = results['tracks']
             show_tracks(tracks)
```
dieser-niko, line 30: Related to: https://github.com/spotipy-dev/spotipy/pull/1068#issuecomment-2089318419
```suggestion
            tracks = sp.playlist_tracks(playlist['id'], fields="items,next")
            show_tracks(tracks)
```

Jack-Dane:
@dieser-niko, thanks for the review, I found the same when investigating the bug, but I didn't know the method you mentioned was available. 

I agree with your proposed changes, the slight change being that `playlist_tracks` is deprecated with the suggestion to use `playlist_items`. Still, it calls the same endpoint mentioned.

dieser-niko:
I guess I missed the deprecation warning, there were just too many prints :P

One thing tho, even though the warning says to use `additional_types=('track',))`, it didn't say to remove the fields parameter. I think it would be better to add it since it was a part of the example before. In this case it should be `fields="items,next"`.

Jack-Dane:
@dieser-niko, yep, my bad, updated to still include the `fields` kwarg.

dieser-niko (approved):

Msoyosa:
Hi Jack, thank you for your plugin for CF7 integration with Odoo. Can I ask how to include the company name and marketing tag, the marketing tag has a Odoo field name  "x_studio_marketing_tag" - many to many, because they are not reflecting in Odoo, both the company name and marketing tag. Thank you.

dieser-niko:
@Msoyosa please visit Jack Danes repository and open an issue there. The topic of your comment does not seem to be related to the Spotipy library and should not be discussed here.

PR #1074, opened 2024-02-22
Supports the use of str in the recommendations function argument and Added Implemented url_to_id function to extract id from url

CureSaba:
# Supports the use of str in the recommendations function argument
## why I did
This is to solve the problem that "," is inserted between characters by .join when str type is specified as a factor of the recommendations function.

# Added url_to_id function to util module You can use this to extract id from url
## why i did
The spotipy document says that URL input can be used, but the official API document says that it cannot be used, so I created a conversion function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added integration tests for audiobook endpoints.
 - Removed `python 2.7` from GitHub Actions CI workflow. Python v2.7 reached end of life support and is no longer supported by Ubuntu 20.04.
 - Removed `python 3.6` from GitHub Actions CI workflow. Ubuntu 20.04 is not available in GitHub Actions for `python 3.6`.
+- Added extra installation step to TUTORIAL.md for required installation packages.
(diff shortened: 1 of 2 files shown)


Mews:
A url to id function would be goated

dieser-niko:
There's already a function that converts URIs/URLs to IDs called `Spotify._get_id`.

But imo the `url_to_id` function from CureSaba would still be useful for programmers, since the already implemented function is intended for internal use within the Spotify module. It's still not perfect though. For example there are other types of IDs like playlists, episodes, albums, artists, etc.
But currently the function would only work for tracks.
There's already a regex string for URIs and URLs called `_regex_spotify_uri` and `_regex_spotify_url`.

Also the parameter `url` is implying that a string is being expected, but the input has to be a list of strings instead.
The variable name `a` should be renamed to something more reasonable and the docstring for the function is outside of the function.

About the implementation in `client.py`. I'm not sure this is a good idea. Checking the Spotify API shows that Spotify specifies the parameter as being of type list, and the name of the parameter (which is plural) also implies that it should be a list. I understand that this change might make things easier, but adding brackets to an argument shouldn't be too hard either.

PR #1087, opened 2024-04-23
Update index.rst(spelling issue)

jackattack825:
Ironic spelling issue in the documents


dieser-niko (approved):
I don't think we need to add this change to CHANGELOG.md.

PR #1096, opened 2024-05-01
Add Ukraine country code

oliversen:


dieser-niko:
I don't think the addition of missing values is necessary, as the only use of `country_code` is in the already deprecated `search_markets` function. If you really need all country codes you can use libraries like [pycountry](https://pypi.org/project/pycountry/).

PR #1100, opened 2024-05-05
Updated _regex_spotify_url to ignore /intl-<countrycode> in Spotify links

dieser-niko:
closes https://github.com/spotipy-dev/spotipy/issues/1071

### DIFF
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -118,7 +118,7 @@ class Spotify(object):
     # instead of a more complex URL parsing function.
     #
     # [1] https://developer.spotify.com/documentation/web-api/#spotify-uris-and-ids
-    _regex_spotify_url = r'^(http[s]?:\/\/)?open.spotify.com\/(?P<type>track|artist|album|playlist|show|episode|user|audiobook)\/(?P<id>[0-9A-Za-z]+)(\?.*)?$'  # noqa: E501
+    _regex_spotify_url = r'^(http[s]?:\/\/)?open.spotify.com\/(intl-\w\w\/)?(?P<type>track|artist|album|playlist|show|episode|user|audiobook)\/(?P<id>[0-9A-Za-z]+)(\?.*)?$'  # noqa: E501
 
     _regex_base62 = r'^[0-9A-Za-z]+$'
 
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -117,7 +117,7 @@ class Spotify(object):
     # pointing to open.spotify.com, so a regex is used to parse them as well,
     # instead of a more complex URL parsing function.
     # Spotify recently added "/intl-<countrycode>" to their links. This change is undocumented.
-    # There is an assumption that the country code uses the ISO 3166-1 alpha-2 standard [2], 
+    # There is an assumption that the country code uses the ISO 3166-1 alpha-2 standard [2],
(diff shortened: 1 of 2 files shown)


dieser-niko:
I just wanted to add you as a reviewer to make sure I wasn't doing anything wrong. In the future, I'll probably merge without you having to review it (unless you'd prefer that).

dieser-niko:
I've come to realise that I'm not allowed to review the pull requst myself and therefore can't merge.

spotipy/client.py
```diff
@@ -118,7 +118,7 @@ class Spotify(object):
     # instead of a more complex URL parsing function.
     #
     # [1] https://developer.spotify.com/documentation/web-api/#spotify-uris-and-ids
-    _regex_spotify_url = r'^(http[s]?:\/\/)?open.spotify.com\/(?P<type>track|artist|album|playlist|show|episode|user|audiobook)\/(?P<id>[0-9A-Za-z]+)(\?.*)?$'  # noqa: E501
+    _regex_spotify_url = r'^(http[s]?:\/\/)?open.spotify.com\/(intl-\w\w\/)?(?P<type>track|artist|album|playlist|show|episode|user|audiobook)\/(?P<id>[0-9A-Za-z]+)(\?.*)?$'  # noqa: E501
```
stephanebruckert, line 121: If you can, let's add as a comment a regex101.com URL so we can quickly check which strings work with the regex. It would kind of serve as a test

stephanebruckert (approved):

spotipy/client.py
dieser-niko: I tried to collect all possible types, but I couldn't find a link with `audiobook` in it. Not even the [Wayback Machine](https://web.archive.org/web/*/https://open.spotify.com/audiobook/*) gave a result.

I found this link here, but doesn't work either:
https://github.com/spotipy-dev/spotipy/blob/070b54f494474e17d3c3e91050d52b3b4a9b2c95/tests/integration/non_user_endpoints/test.py#L60

stephanebruckert: probably fine for now as `intl` links seem to work!

dieser-niko: I'm not sure how to "implement" the comment. The other referenced links seem to be referenced by a description, but there's nothing much to describe about a regex101 link except that it can be used to validate the functionality.

Klairm: Hi, wouldnt a regex like `/intl[\D][a-z]*`  be more simple? Is the one I currently use for my projects where I use spotipy

dieser-niko: This is probably a somewhat objective statement, but I don't see why you think that your regex is simpler than mine. There are only two letters and therefore no fancy `\D` and similar additions are required. Just two `\w` and that's it.

Klairm: > This is probably a somewhat objective statement, but I don't see why you think that your regex is simpler than mine. There are only two letters and therefore no fancy `\D` and similar additions are required. Just two `\w` and that's it.

Yes but what if the internalization code changes in the future or even if theres a country that uses 3 letters?

chaffinched: > Yes but what if the internalization code changes in the future or even if theres a country that uses 3 letters?
Spotify appears to use:
https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2
and not:
https://en.wikipedia.org/wiki/ISO_3166-1_alpha-3

PR #1104, opened 2024-05-09
Updated TUTORIAL.md with an additional instructional step and Troubleshooting Section

dianna-SE:
I added an extra step for installing a package as this part was missing. It might be intuitive but some users may forget and would encounter an error. This step was included in README but not in TUTORIAL. A Troubleshooting section was added to address some common issues with installations. Changes were updated to CHANGELOG.md.


dieser-niko:
I think you managed to commit something that isn't related to the description of your pull request

dianna-SE:
Thanks for the catch! Removed the wrong commits from the PR and made another branch for the other work.

dieser-niko:
I'd recommend doing it the other way around, but that's up to you. As for the changes themselves, they look fine and it makes sense to include some tips in the tutorial.

dieser-niko (approved):
Some more tips would be appreciated, but maybe in the future.

PR #1111, opened 2024-05-20
update playlist_track function

CureSaba:
I included _get_id function into playlist_track function.


dieser-niko:
Hi, this function is already marked as deprecated, so there are no fixes planned for it. Please use `playlist_items` instead (as the deprecation warning already points out)

PR #1116, opened 2024-05-22
Move country codes and remove audiobook functionality

z3nrider:
Make country codes as an importable file in client.py; remove audiobook functionality as it returns an error (possibly being rejected by the Spotify Web API)


dieser-niko:
@stephanebruckert should we remove the country codes list in v3? It only has one use in a deprecated function (`search_markets`). Removing anything related to audiobooks could also be done in v3.

dieser-niko:
Btw, the audiobook feature is only available in the following countries: US, UK, Canada, Ireland, New Zealand and Australia.

But at least we have access to the API documentation: https://developer.spotify.com/documentation/web-api/reference/get-an-audiobook

stephanebruckert:
> @stephanebruckert should we remove the country codes list in v3? It only has one use in a deprecated function (`search_markets`). Removing anything related to audiobooks could also be done in v3.

Yes let's remove `search_markets` in v3 to keep the library straight to the point. We could always create some sort of helper library or move it to the examples repo.

PR #1127, opened 2024-05-31
Remove obsolete `mock` dependency

mgorny:
The package uses `unittest.mock` only, and that one is part of the Python standard library.


stephanebruckert (approved):

mgorny:
Thanks!

PR #1150, opened 2024-07-28
Add Discord badge to README

stephanebruckert:
@dieser-niko and I have set up a Discord server. Just adding the badge from https://github.com/gitlimes/discord-md-badge

Its purpose isn't to replace GitHub issues or existing documentation/examples. Instead, it's for quicker assistance or when creating a GitHub issue might not seem necessary. If a topic is important, we might still ask you to create a GitHub issue for it.

When you join, please read the rules and request any suggestions #server-requests. If you can, set your server username to match your GitHub username so it's easier to link each other to existing contributions.

@plamere if you can join, we will make you an admin as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Add your changes below.
 - Added rate/request limit to FAQ
 - Added custom `urllib3.Retry` class for printing a warning when a rate/request limit is reached.
 - Added `personalized_playlist.py`, `track_recommendations.py`, and `audio_features_analysis.py` to `/examples`.
+- Discord badge in README
 
 ### Fixed
 - Audiobook integration tests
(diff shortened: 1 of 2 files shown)


dieser-niko:
Instead of gitlimes we could also use [shields.io](https://shields.io) instead, just to name an alternative.

dieser-niko:
So it seems that if I'm trying to use shields.io, not only do I need to add the server ID + enable widgets in Discord, but I also need to add an extra invite link. That's one more attribute compared to gitlimes.

stephanebruckert:
Yeah shields.io looks better, I replaced it

dieser-niko (approved):
We should probably mention the server in the next release as well.

PR #1167, opened 2024-11-23
Troubleshooting aid for headless mode

Robin-St:
While trying to use headless mode on a raspberry pi from a Mac I noticed that safari does not properly handle the redirect from the authentication page. This gave that the url in the address field never got updated to the one with the credential token. Testing the same flow and settings in firefox worked directly. 

Added this small note to help other developers potentially avoid to also get stuck for too long on this thing :)


stephanebruckert:
Closing this as it's too vague to add to the code. We need to identify why it failed in Safari and ensure compatibility across all browsers. Feel free to open a new issue with a reproducible example.

PR #1171, opened 2024-11-27
Update docs for `current_user_top_artists`

s2t2:
Update docs to include the maximum value for `limit` parameter of `Client.current_user_top_artists` method. 

See: https://developer.spotify.com/documentation/web-api/reference/get-users-top-artists-and-tracks


dieser-niko:
I think we should add the limits to all docstrings with a limit defined. It's already pretty inconsistent, so this PR might be a good way to fix this. If you want to, you can already start.

stephanebruckert:
> I think we should add the limits to all docstrings with a limit defined. It's already pretty inconsistent, so this PR might be a good way to fix this. If you want to, you can already start.

To clean up PRs, let's merge it as it's already useful, and I've added a comment about it in https://github.com/spotipy-dev/spotipy/issues/652#issuecomment-2585783341

stephanebruckert (approved):

PR #1178, opened 2024-12-03
Formatted strings

### LINKED ISSUE #1159: changing %s to  f-string
Just a small fix of Python 2.x that uses %s in strings instead of f-strings, which is better for Python 3.x.

for example:
```python
logger.error("Can't find artist: %s", artist)
```

should be with python3.x
```python
logger.error(f"Can't find artist: {artist}")
```

dieser-niko:
closes #1159

There are some capitalisations in there as well.
Also, I had to change the playlist in `examples/follow_playlist.py` to a user playlist due to the recent API changes.


dieser-niko:
@haviv1idan you can also check out the PR. I've included some of your changes, but not all. For some, I have added a comment to your PR (#1165)

kampelmuehler:
Why mix `.format()` and f-strings?

docs/index.rst
```diff
@@ -145,7 +145,7 @@ class SpotifyClientCredentials that can be used to authenticate requests like so
     playlists = sp.user_playlists('spotify')
     while playlists:
         for i, playlist in enumerate(playlists['items']):
-            print("%4d %s %s" % (i + 1 + playlists['offset'], playlist['uri'],  playlist['name']))
+            print("{:4d} {} {}".format(i + 1 + playlists['offset'], playlist['uri'],  playlist['name']))
```
haviv1idan, line 148: @dieser-niko I think we need to be consistent in our code. 
Most of files are using f-string instead of format.
You can replace this code to this:
```python
print(f"{i + 1 + playlists['offset']:4d} {playlist['uri']} {playlist['name']}")
```

spotipy/client.py
```diff
@@ -263,8 +263,8 @@ def _internal_call(self, method, url, payload, params):
         if self.language is not None:
             headers["Accept-Language"] = self.language
 
-        logger.debug('Sending %s to %s with Params: %s Headers: %s and Body: %r ',
-                     method, url, args.get("params"), headers, args.get('data'))
+        logger.debug('Sending {} to {} with Params: {} Headers: {} and Body: {!r} '.format(
+                method, url, args.get("params"), headers, args.get('data')))
```
haviv1idan, line 267: The same here. replace format with f-string

spotipy/client.py
```diff
@@ -290,8 +290,9 @@ def _internal_call(self, method, url, payload, params):
                 reason = None
 
             logger.error(
-                'HTTP Error for %s to %s with Params: %s returned %s due to %s',
-                method, url, args.get("params"), response.status_code, msg
+                'HTTP Error for {} to {} with Params: {} returned {} due to {}'.format(
+                    method, url, args.get("params"), response.status_code, msg
```
haviv1idan, line 294: The same here. replace format with f-string

spotipy/oauth2.py
```diff
@@ -218,8 +218,8 @@ def _request_access_token(self):
         )
 
         logger.debug(
-            "sending POST request to %s with Headers: %s and Body: %r",
-            self.OAUTH_TOKEN_URL, headers, payload
+            "Sending POST request to {} with Headers: {} and Body: {!r}".format(
+                self.OAUTH_TOKEN_URL, headers, payload)
```
haviv1idan, line 222: The same here. replace format with f-string

spotipy/oauth2.py
```diff
@@ -511,8 +510,8 @@ def get_access_token(self, code=None, as_dict=True, check_cache=True):
         headers = self._make_authorization_headers()
 
         logger.debug(
-            "sending POST request to %s with Headers: %s and Body: %r",
-            self.OAUTH_TOKEN_URL, headers, payload
+            "Sending POST request to {} with Headers: {} and Body: {!r}".format(
+                self.OAUTH_TOKEN_URL, headers, payload)
```
haviv1idan, line 514: The same here. replace format with f-string

spotipy/oauth2.py
```diff
@@ -541,8 +540,8 @@ def refresh_access_token(self, refresh_token):
...
```
haviv1idan, line 544: The same here. replace format with f-string

spotipy/oauth2.py
```diff
@@ -868,8 +866,8 @@ def get_access_token(self, code=None, check_cache=True):
         headers = {"Content-Type": "application/x-www-form-urlencoded"}
 
         logger.debug(
-            "sending POST request to %s with Headers: %s and Body: %r",
-            self.OAUTH_TOKEN_URL, headers, payload
+            "Sending POST request to {} with Headers: {} and Body: {!r}".format(
+                self.OAUTH_TOKEN_URL, headers, payload)
```
haviv1idan, line 870: The same here. replace format with f-string

spotipy/oauth2.py
```diff
@@ -899,8 +897,8 @@ def refresh_access_token(self, refresh_token):
...
```
haviv1idan, line 901: The same here. replace format with f-string

haviv1idan (changes requested):
@dieser-niko Very nice job. 
Just a few suggestions to keep consistency in code

dieser-niko:
Yeah honestly I don't know what went through my mind. I guess I preferred to use .format() because it would be easier to read, but I agree, consistency would actually be better.

Currently I don't have the time to fix, expect activity in the next month.

docs/index.rst
dieser-niko: I only went through the .py files, looks like I missed that one

haviv1idan (approved):
LGTM 🔥

stephanebruckert (approved):

PR #1184, opened 2025-01-23
Check f-strings

stephanebruckert:


dieser-niko (approved):
Forked it and tried out the workflow myself. Works as expected.

PR #1185, opened 2025-01-24
Convert docstrings to reStructuredText

dieser-niko:
Currently there are two different types of docstrings being used and I only know one of them. Also RTD seems to be able to format RST better than the other format.

Using GitHub Copilot for that job and it seems to shorten a lot of descriptions, even when asked not to.
I'm still going to push these "mistakes" and I'll check afterwards if these changes make sense or not.

### DIFF
--- a/spotipy/cache_handler.py
+++ b/spotipy/cache_handler.py
@@ -48,15 +48,25 @@ class CacheHandler(ABC):
 
     @abstractmethod
     def get_cached_token(self) -> TokenInfo | None:
-        """Get and return a token_info dictionary object."""
+        """
+        Get and return a token_info dictionary object.
+
+        :return: A token_info dictionary object or None if no token is cached.
+        """
 
     @abstractmethod
     def save_token_to_cache(self, token_info: TokenInfo) -> None:
-        """Save a token_info dictionary object to the cache and return None."""
+        """
+        Save a token_info dictionary object to the cache.
+
+        :param token_info: A token_info dictionary object to be cached.
+        """
 
 
 class CacheFileHandler(CacheHandler):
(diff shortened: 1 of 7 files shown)


dieser-niko:
I mostly excluded private methods or getter/setter methods.
@stephanebruckert should all methods have a docstring?

Also I didn't add `:type <var>: <type>` and `:rtype: <type>` because that should be defined by the function/method definition IMO.

dieser-niko:
RTD already looks much better:
Old: https://spotipy.readthedocs.io/en/2.25.0/#spotipy.client.Spotify.artist_albums
New: https://spotipy.readthedocs.io/en/feature-docstring-improvement/#spotipy.client.Spotify.artist_albums

spotipy/cache_handler.py
```diff
@@ -212,29 +271,38 @@ def get_cached_token(self) -> TokenInfo | None:
...
-    """A Cache handler that stores the token info in Memcache using the pymemcache client
+    """
+    A cache handler that stores the token info in Memcache using the pymemcache client.
     """
 
     def __init__(self, memcache, key=None) -> None:
         """
-        Parameters:
-            * memcache: memcache client object provided by pymemcache
-            (https://pymemcache.readthedocs.io/en/latest/getting_started.html)
```
dieser-niko, line 230: Copilot removed the link to pymemcache and honestly, I somewhat agree. If somebody wants to use this CacheHandler with memcache, I'd expect them to know what it is.

dieser-niko (commented):
Went through all files. Content seems to be all right. Copilot did a good job.
Next I'm going to fix the structure.

stephanebruckert (approved):

PR #1187, opened 2025-02-16
fixed Spotify's deprecation of http and localhost

### LINKED ISSUE #1186: Spotify is increasing the security requirements - redirect uri
from spotify's [blog post](https://developer.spotify.com/blog/2025-02-12-increasing-the-security-requirements-for-integrating-with-spotify):

> To improve security for our users, Spotify is planning to remove support for two ways of integrating with Spotify that have been replaced with more secure alternatives. This includes deprecating use of [the implicit grant](https://developer.spotify.com/documentation/web-api/tutorials/implicit-flow) as well as removing support for unencrypted HTTP redirect URIs for your client.
> 
> All clients created from the 9th of April 2025 will have the new rules enforced automatically. Migration of existing clients to an acceptable state must be made by November 2025.

most important thing is that `localhost` will no longer be a supported redirect uri so it must be removed from every docs or readme and changed to `127.0.0.1` and maybe there should be a warning when someone will try to use `localhost` as a redirect uri

also from spotify's blog post:

> Any redirect URI using HTTP will stop being supported, except loopback IP address literals such as http://127.0.0.1 for IPv4 and http://[::1] for IPv6. Any invalid redirect URIs will need to be changed. You can check them under your client’s ‘settings’ tab in the developer console.
> 
> Redirects using a custom scheme will still be supported, but we recommend developers to use HTTPS redirects where possible. For mobile applications, we recommend using Android App Links and iOS Universal Links where possible.
> 
> For example:
> 
> http://www.example.com could be migrated to https://www.example.com
> http://localhost:3000 could be migrated to http://127.0.0.1:3000
> com.example://callback can still be used as before

sandrzejewskipl:
- Added warnings and updated docs due to Spotify's deprecation of HTTP and "localhost" redirect URIs

fixes #1186

### DIFF
--- a/spotipy/oauth2.py
+++ b/spotipy/oauth2.py
@@ -443,6 +443,12 @@ class SpotifyOAuth(SpotifyAuthBase):
         redirect_info = urlparse(self.redirect_uri)
         redirect_host, redirect_port = get_host_port(redirect_info.netloc)
 
+        if redirect_host=='localhost':
+            logger.warning("Using 'localhost' as a redirect URI is being deprecated. Use a loopback IP address such as 127.0.0.1 to ensure your app remains functional.")
+
+        if redirect_info.scheme == "http" and not redirect_host in ("127.0.0.1", "localhost"):
+            logger.warning("Redirect URIs using HTTP are being deprecated. To ensure your app remains functional, use HTTPS instead.")
+
         if open_browser is None:
             open_browser = self.open_browser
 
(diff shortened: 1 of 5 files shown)


stephanebruckert:
Hey, just saw that, thanks for doing the changes!

For the tests to pass I might have to update the redirect URI in the Spotify API app. I will check and let you know

stephanebruckert:
I added http://127.0.0.1:8080, but doesn't look like it helps
I then removed http://localhost:8080, also tried to re-generate the app credentials, same thing.
Currently both URIs are set

stephanebruckert:
Oh got it, because you are a first-time contributor you don't have access to GHA secrets https://github.com/spotipy-dev/spotipy/actions/runs/13355321896/job/37440513555#step:5:5

spotipy/oauth2.py
```diff
@@ -743,6 +754,17 @@ def _get_auth_response(self, open_browser=None):
         if open_browser is None:
             open_browser = self.open_browser
 
+        if redirect_host == 'localhost':
+            logger.warning(
+                "Using 'localhost' as a redirect URI is being deprecated. "
+                "Use a loopback IP address such as 127.0.0.1 "
+                "to ensure your app remains functional.")
+
+        if redirect_info.scheme == "http" and not redirect_host not in ("127.0.0.1", "localhost"):
```
stephanebruckert, line 763: and not redirect_host not in

did you mean the double negation here?

sandrzejewskipl: no, i changed in to not in and forgot to remove non after and, sorry

stephanebruckert (approved):

PR #1193, opened 2025-04-19
add additional_types to current_user_playing_track()

sibbi77:
The new Spotify audiobook feature plays as episodes not as tracks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ Add your changes below.
 
 ### Added
 
+- Adds `additional_types` parameter to retrieve currently playing episode
+
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


dieser-niko:
Nice catch. I think your change is good enough for V2, but I think we should rename the function to `current_user_playing` for the upcoming V3. I don't really see the point of checking only tracks.

I'll look into why the tests are failing again later today.
@stephanebruckert maybe you can have a look at it if you have some free time.

stephanebruckert:
@dieser-niko I've updated the integration tests workflow so that it passes the secrets to runs from fork. All green now!

stephanebruckert (approved):

PR #1203, opened 2025-05-26
Fix indefinite hang on invalid PKCE client_id

AtomikkuLabs:
Fixes indefinite hang when an invalid client_id is put into SpotifyPKCE and get_access_token is called. A preflight has been made to test out the client_id beforehand. Not sure if this is the most valid approach but it seems to work and gives a clear error.

PKCE issue was noted by me [here](https://github.com/spotipy-dev/spotipy/issues/957#issuecomment-2826480651)

### DIFF
--- a/spotipy/oauth2.py
+++ b/spotipy/oauth2.py
@@ -875,6 +875,27 @@ class SpotifyPKCE(SpotifyAuthBase):
         if self.code_verifier is None or self.code_challenge is None:
             self.get_pkce_handshake_parameters()
 
+        preflight_payload = {
+            "client_id": self.client_id,
+            "grant_type": "authorization_code",
+            "code":          "invalid_code",
+            "redirect_uri": self.redirect_uri,
+            "code_verifier": self.code_verifier,
+        }
+
+        preflight_resp = self._session.post(
+            self.OAUTH_TOKEN_URL,
+            data=preflight_payload,
+            headers={"Content-Type": "application/x-www-form-urlencoded"},
+            timeout=self.requests_timeout,
+            proxies=self.proxies
+        )
+
+        if preflight_resp.status_code == 400:
+            body = preflight_resp.json()
+            if body.get("error") == "invalid_client":
+                raise SpotifyOauthError("Invalid client_id")
+
         payload = {
             "client_id": self.client_id,
             "grant_type": "authorization_code",


dieser-niko:
Sorry for the late reply. But I have to say, I really like it!

Just one thing. We won't be able to verify the redirect_uri, are we?

BigAtomikku:
> Sorry for the late reply. But I have to say, I really like it!
> 
> Just one thing. We won't be able to verify the redirect_uri, are we?

Thanks for the reply @dieser-niko. An invalid redirect_uri will cause the preflight to fail as well but that does not matter as we only raise an invalid client_id error and prevent the webpage from opening (therefore freezing the program) when we get a `'invalid_client'` error from the preflights response. If the redirect_uri is wrong it will produce a generic `'invalid_grant'` response which will cause it to just continue and open the webpage as normal, producing "INVALID_CLIENT: Invalid redirect URI" and freezing the program. I do not see this as an issue as developers should be putting a correct redirect_uri in statically. Since programs like my own need to allow a user to input their own client_id which a user may put in garbage, it needs to not freeze the program like it currently does without my preflight.

AtomikkuLabs:
Was wondering if there was any update about merging this pull request or things youd need me to change?

stephanebruckert:
Hey @AtomikkuLabs, thanks for this and apologies for reviewing it late!

Discussed with @dieser-niko and we figured we probably want to avoid doing a pre-flight check (extra request) on all calls including production. An `invalid_client` is a development/config issue which should be fixed directly by the developer rather than by adding extra code for it.

That being said we can have the code give better feedback and improve the developer experience around:

- Ctrl + C doesn't work at all on Windows as mentioned here https://github.com/spotipy-dev/spotipy/issues/957#issuecomment-2826664334.

  <details>
  <summary>possible solution</summary>

  > It's a known CPython limitation on Windows (open issue https://github.com/python/cpython/issues/85609).
  >
  > On Linux: Ctrl+C sends SIGINT which interrupts the blocking select() call inside server.handle_request(), allowing Python to raise KeyboardInterrupt.
  > 
  > On Windows: Ctrl+C spawns a new thread to handle the console event. The main thread stays blocked inside the Winsock select() call and is never interrupted.
  >  Python's signal handler sets a flag, but the main thread never gets a chance to check it because it's stuck in the C-level blocking call.
  > 
  > The fix would look like:
  > 
  >     server.timeout = 1.0  # return every second
  >     while True:
  >         server.handle_request()
  >         if server.auth_code or server.error:
  >             break
  >         # Python VM checks signals here, so Ctrl+C works on Windows too
  </details>

 - Currently hangs indefinitely when `invalid_client` or no action taken, a 5-min timeout would be fair (may be able to use the same code as above).
 - The waiting message could be improved to hint at potential issues:

       Waiting for authentication in your browser... If nothing happens, please verify your `client_id` and `redirect_uri`.

- This should work for both OAuth and PKCE, as both require the user to complete an action in the browser.

Let me know if you prefer to close this PR, improve it or if you'd like some help! Thanks a lot

PR #1207, opened 2025-06-26
V3

thedevappsecguy:


thedevappsecguy:
@stephanebruckert I see this require approval now for the flow to trigger not like the case earlier.
https://github.com/spotipy-dev/spotipy/pull/1206 I suggest better to check and remove the branches where all the issue exists or fix the flow in all the branches. Thank you !

thedevappsecguy:
@stephanebruckert Thanks .My bad I tested and it did not work as expected.So closing the PR.

Issue #1217, opened 2025-11-06
Spotify no longer allows `localhost`


zlonghofer:
Per [this article](https://developer.spotify.com/documentation/web-api/concepts/redirect_uri#:~:text=localhost%20is%20not%20allowed%20as%20redirect%20URI.), Spotify won't allow `localhost` as a redirect URI and expect all clients to migrate by this month. All references to `localhost` in the documentation should be updated to reflect `127.0.0.1` instead.

PR #1218, opened 2025-11-06
Remove localhost from documentation

zlonghofer:
This PR resolves issue #1217 by updating the documentation.


dieser-niko (commented):
Thanks for noticing!
Seems like you didn't set up CI/CD, but that's alright. I'm just somewhat busy this weekend, so I'll need to check it out next week.

The thing is, I don't know if the v3 branch is the correct one to work on right now and I won't be able to figure that one out. I hope you can understand that.

PR #1220, opened 2025-12-02
Add type hinting to function args; remove support for deprecated versions 3.8 and 3.9

bvandercar-vt:
- Add Python types to function args. Typing is always better as it can catch errors before a script is ran.
- Remove support for 3.8 and 3.9, as these are now deprecated. Allows us to use pipes in types (ie `str | None` instead of `Optional[str]`
- Integration tests passing: https://github.com/bvandercar-vt/spotipy/actions/workflows/integration_tests.yml


bvandercar-vt:
@dieser-niko wanna take a look?

sakgoyal:
3.8 and 3.9 are now deprecated. can this be revisited please?

dieser-niko:
I haven't forgotten about your PR, I just need to find the time for it.

PR #1222, opened 2025-12-22
fix: dont override verify attribute when requesting access token

sigma67:
This call is the only one that overrides the `_session.verify` which can be passed in via the arguments of `SpotifyClientCredentials`.

This makes it impossible to actually disable cert verification or set a custom certificate chain (i.e. in a corporate network).

Hence this line should be removed and use the default settings instead, which are provided by the caller.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ Add your changes below.
 
 ### Fixed
 
--  Don't override verify attribute when requesting access token in SpotifyClientCredentials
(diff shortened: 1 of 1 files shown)


dieser-niko:
Sounds good. Also would you mind setting up your CI/CD? I just want to make sure that your checks can succeed.

Just make sure to add the following variable secrets to your GitHub project settings:
- `SPOTIPY_CLIENT_ID`
- `SPOTIPY_CLIENT_SECRET`

Plus add an entry to the `CHANGELOG.md` file

@stephanebruckert anything against it?

PR #1225, opened 2025-12-23
Warning on Spotify URI RegEx fail

### LINKED ISSUE #1221: Local Tracks Not Recognized as URI
**Describe the bug**
The client.py _is_uri() method is throwing an exception when the track is a local file.

When the URI is something "normal", like "spotify:track:7hK9jnESeCfc5fIHH76jKh", the _is_uri() method recognizes it as valid via the regex match.

However, when the URI is for a local track, like "spotify:local:Toot+Sweet:Celtic+Christmas:Christmas+Medley:248", the _is_uri() regex match fails. It then calls _get_id("track", "spotify:local:Toot+Sweet:Celtic+Christmas:Christmas+Medley:248"), and none of the matches work in that function and it eventually throws an Unsupported URL SpotifyException (currently on line 2139).

**Your code**
I have a list of URIs that I've retrieved via client.playlist_items().

When I call the following function, passing that list of URIs, that's when the exception is thrown.

self.api.playlist_replace_items(playlist_id, uris[:100])

spotipy.exceptions.SpotifyException: http status: 400, code: -1 - Unsupported URL / URI., reason: None

**Expected behavior**
The function works without throwing an exception

**Output**
(self.api is an instance of the Spotify class)
(playlist is an object with an **id** attribute)

  File "D:\Documents\Python\spotify-backup\spotify_api.py", line 94, in update_tracks
    self.api.playlist_replace_items(playlist.id, uris[:chunk])  # Replace playlist's first 100 items (limit of Spotify)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python\Lib\site-packages\spotipy\client.py", line 1187, in playlist_replace_items
    ftracks = [self._get_uri("track", tid) for tid in items]
               ~~~~~~~~~~~~~^^^^^^^^^^^^^^
  File "C:\Python\Lib\site-packages\spotipy\client.py", line 2145, in _get_uri
    return "spotify:" + type + ":" + self._get_id(type, id)
                                     ~~~~~~~~~~~~^^^^^^^^^^
  File "C:\Python\Lib\site-packages\spotipy\client.py", line 2139, in _get_id
    raise SpotifyException(400, -1, "Unsupported URL / URI.")
spotipy.exceptions.SpotifyException: http status: 400, code: -1 - Unsupported URL / URI., reason: None

**Environment:**
 - OS: Windows 11
 - Python version: 3.14.0
 - spotipy version: 2.25.2
 - your IDE: PyCharm

**Additional context**
I believe the fix would be to modify the _regex_spotify_uri regex pattern to account for the local file syntax, although I'm not sure what the Spotify syntax spec for local files is.

It looks like it might be spotify:local:artist_name:album_name:track_name:number.
My local file is:
**Artist:** Toot Sweet
**Album:** Celtic Christmas
**Title:** Christmas Medley
**Filename:** "Christmas Medley.mp3"

and the Spotify URI for it is "spotify:local:Toot+Sweet:Celtic+Christmas:Christmas+Medley:248". I have no idea what the '248' indicates.

FYI, the **type** parameter being passed is "track".

dieser-niko:
Closes #1221

While implementing, I noticed that such a warning would probably be more disruptive than helpful.
Currently there are multiple methods that *should* accept both tracks and episodes, but the regex only checks for tracks.

Any thoughts?

### DIFF
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -2140,6 +2140,7 @@ class Spotify:
 
     def _get_uri(self, type, id):
         if self._is_uri(id):
+            warnings.warn(f"{repr(id)} did not match the Spotify URI RegEx.", UserWarning)
             return id
         else:
             return "spotify:" + type + ":" + self._get_id(type, id)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ Add your changes below.
 
 ### Added
 
+- Added a warning if the URI regex did not match the input.
+
 ### Fixed
 
 ### Removed


spotipy/client.py
```diff
@@ -2140,6 +2140,7 @@ def _get_id(self, type, id):
 
     def _get_uri(self, type, id):
         if self._is_uri(id):
+            warnings.warn(f"{repr(id)} did not match the Spotify URI RegEx.", UserWarning)
```
tinomen: Can we be more explicit that it is the Spotify API, since this is quite possible from the spotify apps. I know this is an API library and should be assumed given that context, but ... people.

`did not match the Spotify API URI RegEx.`

tinomen (commented):
Thank you for adding this message

spotipy/client.py
dieser-niko: What would you want in there? Not sure what kind of information we can throw into that message

vinegreet: He just suggests to add 'API' before 'URI'

PR #1228, opened 2026-02-18
Fix: Web API Changes of February 2026

tonarec:
Closes #1227 

This PR fix the Spotiy client to comply with the new changes in the API from Spotify, see the [Web API Changelog](https://developer.spotify.com/documentation/web-api/references/changes/february-2026)

It introduces a change in the way the tracks/episodes (the so named _items_) are handled by Spotify.
In addition, some of the endpoints are now deprecated, and most of the enpoints related to the user's content moved to the **/me/library** endpoint to use URIs only.

Introduced in this PR:
- Updated `/tracks` endpoints to `/items`
- Switching IDs to URIs to use the `/me/library` endpoint [here](https://developer.spotify.com/documentation/web-api/reference/save-library-items)
- Fixed playlist limit to 50 according to the API for `/playlists/{playlist_id}/items` [here](https://developer.spotify.com/documentation/web-api/reference/get-playlists-items)
- Added warnings for deprecated methods

PR #1232, opened 2026-03-11
Reduce `artist_album` limit

### LINKED ISSUE #1231: Invalid limit for `artist_albums`
**Describe the bug**
The default limit for `artist_albums` is no longer valid.

**Your code**
```python
client = ...
client.artist_albums('4s3cCLEcBhVUWLNDFJjRL1')
```

**Expected behavior**
Should return albums

**Output**
```
HTTP Error for GET to https://api.spotify.com/v1/artists/4s3cCLEcBhVUWLNDFJjRL1/albums with Params: {'include_groups': None, 'country': None, 'limit': 15, 'offset': 0} returned 400 due to Invalid limit
```

**Environment:**
 - OS: Linux
 - Python version 3.12.3
 - spotipy version 2.26.0

**Additional context**
From my experimentation, 10 is the highest allowed limit. Current default is 20.

fabi321:
Reduces the limit so that a query called without explicit limit is valid with the new changes to the spotify API.

Fixes #1231

### DIFF
--- a/spotipy/client.py
+++ b/spotipy/client.py
@@ -400,7 +400,7 @@ class Spotify:
         return self._get("artists/?ids=" + ",".join(tlist))
 
     def artist_albums(
-        self, artist_id, album_type=None, include_groups=None, country=None, limit=20, offset=0
+        self, artist_id, album_type=None, include_groups=None, country=None, limit=10, offset=0
     ):
         """ Get Spotify catalog information about an artist's albums
 


dieser-niko (approved):
I went through all limits again just to make sure we don't have similar issues and this one seems to be the only one over the limit.

The only "wrong" limits are lower than the ones in the API, however I think it was the default value, meaning Spotify recommends that limit. Spotipy doesn't stop a user from going over the "limit" as well, so we'll keep it like this.

I honestly doubt that this one was missed by PR #1228, maybe Spotify overlooked that endpoint and changed it afterwards...

dieser-niko:
We won't release this change right away, as this isn't a critical bug or anything, only a default value.

